PR #4, opened 2021-05-19
RFC 0004 - Standardize on CloudEvents when migrating SocketProxy

paulswartz:


rfcs/0004-socket-proxy-ocs-cloudevents.md
```diff
@@ -0,0 +1,134 @@
...
+An example TMOV message, as received from the OCS on 2020-02-11: 
+
+`9529,TMOV,02:00:40,G,10230,42.336251428162655,-71.14902308252195,1,831,25,  3885,gpsci,3885,0.00,254.19,3,42.336251428162655,-71.14902308252195`
+
+The same TMOV message, as a CloudEvent representing the raw message (encoded in JSON): 
+
+```json
+{ 
+  "specversion": "1.0", 
+  "type": "com.mbta.ocs.raw_message",
```
losvedir: Below you mention that if the format changes, we can do a `.v2` or `.[date]`. Should we pre-emptively label this with a version or with a date already? It does add a bit more data to the message, but it also makes it easier to know how to update it in the future if we need to.

You also mention that this provides a schema. Do you have any ideas about where we might centrally document all our CloudEvents schemas, or if we need to or should?

losvedir:
This looks good to me (unsurprisingly, since I already reviewed the draft)! Thanks for taking our RFC process for its maiden voyage.

I think the next step in the process is for me to move this into "Final Comment Period" with a disposition. At this point, my disposition is "accept", but I think I'll hold off on formally doing that until others have had a chance to look at this, if they want, and until its partner RFC about using Kinesis is opened, since I think they are somewhat related.

rfcs/0004-socket-proxy-ocs-cloudevents.md
paulswartz: - I don't love the idea of starting with the version/date in there, but I could be convinced.
- My thought is that we have an `mbta/schemas` repository where the schemas live keyed by their type.

losvedir:
Now that #5 has been published and reviewed, I'm moving this one as well to final comment period with disposition to accept.

losvedir:
Okay, I'm ready to merge! Can you push up a commit changing the status to "accepted" and putting it in the accepted folder?

PR #5, opened 2021-06-02
RFC 0005 - Send OCS messages as JSON-formatted CloudEvents to AWS Kinesis

paulswartz:


rfcs/0000-socket-proxy-kinesis-json.md
```diff
@@ -0,0 +1,168 @@
...
+
+# Motivation
+
+Currently, SocketProxy proxies OCS message to each RTR instance running in AWS. This method of sending Occupancy Control System (OCS) messages has several issues we are trying to address:
+- insecure: the messages travel unencrypted over the public internet
+- insecure: anyone from the MBTA network can forge OCS messages
+- unreliable: if an RTR instance isn't running (crashed, or during a restart/redeploy) when a message comes in, that message is lost
+- unreliable: if the production RTR instance isn't running, then the log of messages written to S3 (and used by OPMI) will not have those messages
+- missing metadata: the messages do not include the date they were sent, or the time the messages were received by SocketProxy 
+
```
losvedir: Also it's wasteful of network bandwidth: currently we send out 4 duplicate streams of OCS messages to the 4 AWS instances.

rfcs/0000-socket-proxy-kinesis-json.md
```diff
@@ -0,0 +1,168 @@
...
+To address these issues, we'll update SocketProxy to write the OCS messages (along with additional metadata) as JSON CloudEvent messages to an AWS Kinesis stream. The partition key will be based on the type of message and the line.
```
losvedir: I think we might want to simply retire Socket Proxy and do a new application here. I'd say the _majority_ of Socket Proxy is about broadcasting one listening socket onto 4 sending sockets that it maintains and manages the state of. In addition, the name no longer would make sense.

losvedir: Also, regarding the partition key: is that necessary? I don't know that we have enough data to warrant a partition, and doing so means the app will have to know more about the underlying message, to parse out at least what kind of message it is, vs. simply buffering and watching for EOT.

Socket Proxy does listen to 4 incoming sockets, which correspond to a kind of message, but I don't know that it *knows* anything about those sockets or where they come from, in order to infer anything from it. Though it could sniff the first message it sees and then just maintain that state for that socket.

Lastly, I think we had decided to split RFCs with hard-breaks, so that we could have somewhat more targeted comments.

rfcs/0000-socket-proxy-kinesis-json.md
```diff
@@ -0,0 +1,168 @@
...
+- Corner cases are dissected by example.
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work. -->
+
+Currently, SocketProxy has a Receiver GenServer responsible for each incoming connection from OCS. Receiver then spawns a Forwarder GenServer to send the data to each of the destinations (currently, the four RTR instances). After this RFC, SocketProxy will have a KinesisReceiver which is responsible for writing records to a configured Kinesis stream:
+
+```
+SOCKET_PROXY_LISTEN_PORT=8000 SOCKET_PROXY_KINESIS_STREAM=ctd-ocs-raw-messages mix run --no-halt
+```
+
+The maximum size of a record (including the partition key) is 1MB, and record storage is charged in units of 25KB. Each shard supports writing 1MB or 1000 records per second For reading, each shard can support 5 read transactions per second, and up to 10,000 records per transaction, up to a total of 2MB per second. One (1) shard should be sufficient, but we will provision two (2) shards. While one shard should be sufficient, having multiple shards ensures that consumers properly handle multiple shards from the start.
```
losvedir: Typo: `1000 records per second For reading`.

rfcs/0000-socket-proxy-kinesis-json.md
```diff
@@ -0,0 +1,168 @@
...
+
+Running Kinesis locally is easy enough with tools like LocalStack or Kinesialite, but more involved than running a TCP proxy. This might require some re-working of how the RTR integration system works when pushing logged data into a local instance for testing.
+
+# Rationale and alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+
+AWS has tools to proxy messages from Kinesis into Kafka (and vis versa) so if we need to migrate in the future that is possible. If we had a lot more operational staff (or more experience with Kafka) this might point to a different decision, but I believe that Kinesis gets us the value we need with the minimum of cost and operational overhead.
```
losvedir: `vice versa`

rfcs/0000-socket-proxy-kinesis-json.md
losvedir: How does back-pressure work? If we exceed those limits, do messages build up, or do they get dropped by Kinesis? We get a steady state of ~5-10 messages/second, but it can spike upon a socket reconnecting. And at 2am we can peak >500 messages in one second (see [here](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Drtr-prod%20NOT%20parsed_ocs_message%20(TMOV%20OR%20TSCH)%20%7C%20bin%20_time%20span%3D1s%20%7C%20stats%20count%20by%20_time&display.page.search.mode=smart&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1622779200&latest=1622791155&display.page.search.tab=visualizations&display.general.type=visualizations&sid=1622830777.3000430)).

On the reading side, is RTR effectively switching to a polling model, say, 1/sec from the push model we have now? The 4 environments would be querying 1/sec, that's at the limits of what Kinesis can support, right? That feels a little on the low side.

And would it make sense for the Socket Proxy replacement to use Broadway/GenStage? I like the simplicity of a bunch cowboy listeners just pushing messages when they get them, but it depends on backpressure I suppose. If Kinesis drops messages then I'd prefer the local app to buffer them, I think.

rfcs/0000-socket-proxy-kinesis-json.md
```diff
@@ -0,0 +1,168 @@
...
+
+By default, records are kept for 24 hours. This can be extended up to 7 days for an extended data retention stream, and 365 days for a long-term data retention stream.
+
+This leads to a few possibilities for managing the state of the stream in RTR.
+
+1. [discouraged] Always read from the latest message in the shard. This replicates the existing behavior when receiving messages directly from SocketProxy. However, this results in skipping messages during restarts.
+2. [encouraged] The current sequence number for each shard is recorded as a part of the RTR state file in S3. When an instance restarts, it can continue from the last message it received.
+3. [encouraged] Using the Kinesis Client Library (KCL) to maintain the current sequence number in DynamoDB. While KCL is a Java application, the Multi-Lang Daemon provides access to other languages include Elixir.
```
losvedir: I see that an `ex-aws/ex_aws_kinesis` exists. How does this relate to KCL?

rfcs/0000-socket-proxy-kinesis-json.md
```diff
@@ -0,0 +1,168 @@
...
+- I expect that the specific implementation of state management will be resolved during development, rather than as a part of this RFC.
+- Some of the specific names (the name of the CloudEvent message, the name of the Kinesis stream) may change, either based on RFC feedback or during development.
+
+## Batching
+There are options for batching messages into a single Kinesis record, which could potentially increase throughput at the cost of some additional complexity.
+
+- a new CloudEvent message type like `com.mbta.ocs.raw-message-batch`
+- the Kinesis Producer Library (KPL) provides a Protobuf-based format for batching, which is also understood by the KCL
+
+For example, during the TSCH load in the morning, the peak traffic is ~400 messages/s. If we batched them even at 1 second granularity, that would reduce the number of messages dramatically.
```
losvedir: Ah, so this would be a good way to handle that spike (or the socket re-connection spike I mentioned). But then... with all this batching and polling are we introducing latency?

Previously OCS -> RTR is a push event which RTR responds to immediately, though it only writes the realtime data 1/sec, so data is going to be at most 1/sec out of date. But now with a 1 second batch on opstech, and a 1 second poll in RTR, are we looking at 3 seconds of delay?

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: The reason I kept this in SocketProxy (even if the name is bad for a time) is so that we can run both the Kinesis-based RTR instances and the current TCP instances in parallel during development. wdyt?

The partition key is required, but we could set it to a random string each time if we didn't care about making sure an ordering is preserved. Or we could use something like `host:port` of the connecting socket, so all the messages from a given connection are routed to the same shard?

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: Kinesis does not drop messages: readers can get behind if they don't process messages fast enough. (Thank you for the Splunk query: I was trying to do that myself and couldn't get it to work).

The 5 reads / second might be raise-able: I'll look into it. Even if it's not, however, RTR should be able to fetch all the messages from the last second in one read even during peak times.

It definitely makes sense for the readers to use Braodway/GenState (the `kcl_ex` implementation of the Kinesis Client Library uses Broadway under the hood). But for the writer, it gets messages at the rate that OCS sends them to us: how much is it possible to give them backpressure?

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: If it's possible to dig into the messages a bit, maybe only some message types get batched? If we know which messages are TMOVs and need to go immediately, we can handle them separately from TSCH which can get bundled.

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: `ex_aws_kinesis` is the wrapper around the Kinesis REST API.

KCL and `kcl_ex` provide some features on top of that:
- shard leases, to distribute work between different running instances of an application
- checkpointing, so an application knows the last message it processed in each shard

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: Digging in, it looks like TMOVs peak at 50/s. It's the TSCH which peak at 400/s but could also be batched effectively. If the polling is done in batches of 1/s, could code currently in RTR to push GTFS-RT updates every second be removed? Then it would go as fast as the polling rate and we'd limit the delay. I'll also do some local testing to see what the latency looks like with 1/s polling rate: since messages are available immediately after writing, the average should be lower than that.

The 5/s/shard read rate isn't raise-able, but there's another option if it turns out that's not enough. [Enhanced fan-out consumers](https://docs.aws.amazon.com/streams/latest/dev/building-enhanced-consumers-api.html) 1) each one gets a dedicated 2MB/s for each shard 2) messages are pushed via HTTP/2 rather than needing to be polled. The disadvantage is that it costs more: enhanced fan out is charged by both consumer/shard/hour and GB retrieved, and the consumers need to be registered ahead of time.

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: I did some testing against Kinesalite. Running at real speed, the average delay is basically zero. Even writing as fast as possible with one thread (500 records per write), the average delay was less than 500ms. I'll try to get some real numbers next week.

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: Looks like AWS also recommends using enhanced fan-out here:
Screen Shot 2021-06-07 at 9 58 06 AM

I'll also look into that this week.

paulswartz: This does work, but we'll need to write some of our own code to support it.

rfcs/0000-socket-proxy-kinesis-json.md
losvedir: > The reason I kept this in SocketProxy (even if the name is bad for a time) is so that we can run both the Kinesis-based RTR instances and the current TCP instances in parallel during development. wdyt?

This will mean repeated deploys of in-development branches to effectively "prod" Socket Proxy. We only have one real environment because the TCP sockets on the OCC end are configured to send to just a single IP/port.

However, one development flow we've used before with success is to have *another* separate app running on some port on opstech3, and having Socket Proxy proxy to `127.0.0.1:[that port]`.

> The partition key is required, but we could set it to a random string each time if we didn't care about making sure an ordering is preserved. Or we could use something like host:port of the connecting socket, so all the messages from a given connection are routed to the same shard?

Yep, I like the idea of `host:port` like in the RFC now. I'm not sure of the semantics of required ordering in RTR, but what we've effectively been using thus far is ordered-per-socket, with sockets interleaved. So `host:port` should maintain that.

Do listeners need to specify the shard? I think the host and/or port can change over time. And would the host and port be considered sensitive information?

rfcs/0000-socket-proxy-kinesis-json.md
losvedir: > But for the writer, it gets messages at the rate that OCS sends them to us: how much is it possible to give them backpressure?

Heh, there's a limited capacity to apply backpressure, but at some point it causes things to 💥 in the OCC (ask me how I know). We should definitely architect to drain incoming messages as quickly as possible.

Well, not listening at all won't cause any issues there, but accepting the TCP connection and then not reading from it fast enough can be a problem. So I suppose if necessary we _could_ close the connection if we were falling behind. Then when we reconnect we'd get some of the interim messages, but I don't think OCC will buffer indefinitely.

> If the polling is done in batches of 1/s, could code currently in RTR to push GTFS-RT updates every second be removed?

Yep. It might be a little bit of work to do so. At the moment, receiving messages is async from processing the GTFS-rt file, and I don't know to what extent the test suite (and in particular integration tests) rely on that. I think it would simplify things overall, though.

Regarding shards and fan-outs, I'll leave that up to you. :) But it sounds like it _can_ handle whatever we need.

losvedir:
Looks good to me! I unresolved some comments for visibility, but basically I'm 👍 on marking this ready for review (with tweak about using a separate app that socket proxy proxies to locally, as optional).

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: - good point on the redeploying SocketProxy: I'll update to indicate that this will be a new application, and we'll update the SocketProxy configuration to point at it.
- Listeners do need to specify the shard but that's true regardless of the partition key. If we're concerned about the host being part of the data, then we can hash it on our end, and use the hashed value as the partition key. That maintains the same ordering, without exposing the host in the data.

rfcs/0000-socket-proxy-kinesis-json.md
paulswartz: I think we can have the KinesisProxy write data as fast as possible. We know that RTR can currently process the data fast enough, and this shouldn't change that.

0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,306 @@
...
+- Asana task: [Secure connection between socket_proxy and rtr](https://app.asana.com/0/881264583703207/1200290609745412/f)
+- Status: Proposed
+
+# Summary
+
+Send OCS messages as JSON-formatted CloudEvents to AWS Kinesis
+
+# Motivation
+
+Currently, SocketProxy proxies OCS message to each RTR instance running in
```
digitalcora: typo: "OCS messages"

0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,306 @@
...
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.  The section should return to the
+examples given in the previous section, and explain more fully how the
+detailed proposal makes those examples work. -->
+
+Currently, SocketProxy has a Receiver GenServer responsible for each incoming
+connection from OCS. Receiver then spawns a Forwarder GenServer to send the
+data to each of the destinations (currently, the four RTR instances). After
+this RFC, a new KinesisProxy which is responsible for
+writing records to a configured Kinesis stream:
```
digitalcora: I think there's a word/phrase missing here: "a new KinesisProxy" ...will be created? Also, is the implication that this is a new Elixir project? A new app within an existing Elixir project? "KinesisProxy" is mentioned in many places but I don't see a concrete explanation of what it would be.

0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,306 @@
...
+   current sequence number in DynamoDB. While KCL is a Java application, the
+   Multi-Lang Daemon provides access to other languages include Elixir. There
+   is also a pure-Elixir implementation. KCL provides functionality on top of
+   the Kinesis and DynamoDB APIs: shard leases (so only one application
+   instance is reading from a shard) and checkpointing (so the application
+   knows which messages have been seen).
+4. [potentially easier, more expensive] RTR stops maintaining state, and
+   starts reading the Kinesis stream from the start of service whenever it
+   starts. This would require increasing the default Kinesis storage above
+   the default of 24 hours, and absorbing some additional cost to use
```
digitalcora: > This would require increasing the default Kinesis storage above the default of 24 hours

I'm curious why, given the start of service is never more than 24 hours ago.

0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,306 @@
...
+   extended data retention.
+
+KinesisProxy will write JSON-encoded CloudEvents to the Kinesis
+stream. JSON-encoded CloudEvents are somewhat larger than a binary encoding,
+but still easily fit within the limits of a Kinesis record. They are also
+human-readable, aiding in debugging.
+
+The partition key will be a hash of the TCP source `host:port`, along with a
+random salt. The random salt is to ensure that it's not possible to go from
```
digitalcora: Is the random salt generated for a specific TCP connection at the time it's established? Just once when the proxy instance starts up?

0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,306 @@
...
+
+# Rationale and alternatives
+
+<!-- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this? -->
+
+AWS has tools to proxy messages from Kinesis into Kafka (and vice versa) so if
+we need to migrate in the future that is possible. If we had a lot more
+operational staff (or more experience with Kafka) this might point to a
```
digitalcora: Do we have any more experience with Kinesis than we do with Kafka, at this point, or is it a question of Kafka being fundamentally more complicated to set up and use? The existence of MSK seems to negate the concern about needing more ops folks to run Kafka.

0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,306 @@
...
+- vendor-neutral: Kafka can run on any cloud or on-premises
+- pricing is based on cluster size and storage, so not as closely tied to the
+  number of events
+- messages have additional metadata, requiring slightly less overhead when
+  using CloudEvents
+- Being used by Cubic for Fare Transformation in Azure
+
+### Disadvantages
+- MSK cannot currently resize a cluster, so we either need to pay more up
+  front, or pay an additional migration cost as we scale up
```
digitalcora: Is "scaling up" much of a concern within the scope of this RFC? I'd expect OCS message volume to be roughly constant over time.

digitalcora (commented):
The bar is being set high for future RFCs 😄 I had some questions, but nothing major.

0005-kinesis-proxy-json.md
losvedir: I'm guessing this is the new Elixir app which will replace SocketProxy. You might not have recognized it as an app name by Paul because it doesn't have some cutesy name like... `Tricycle`, with many levels of meaning. Probably need to open a new RFC to properly bikeshed the name. :)

paulswartz: Now I kind of want to call it Tricycle :)

paulswartz:
@digitalcora thank you for the good feedback! 3272f22 should address it.

digitalcora:
@paulswartz Thanks, I resolved everything that was resolved, though there's one comment left that got marked "outdated" despite the line not changing (?). Good clarifications, and the cost breakdown is interesting as well.

rfcs/0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,320 @@
...
+- insecure: anyone from the MBTA network can forge OCS messages
+- unreliable: if an RTR instance isn't running (crashed, or during a
+  restart/redeploy) when a message comes in, that message is lost
+- unreliable: if the production RTR instance isn't running, then the log of
+  messages written to S3 (and used by OPMI) will not have those messages
+- wasteful: we send each message four separate times to four RTR instances
+- missing metadata: the messages do not include the date they were sent, or
+  the time the messages were received by SocketProxy 
+
+To address these issues, we'll create KinesisProxy, a new Elixir application,
```
losvedir: What about a more descriptive name? I'm thinking something like `OCSMessageProxy` or something to that effect. Kinesis is something of an implementation detail, and I think it would be helpful to 1) have the windows service clearly map to the underlying elixir app, and 2) have that windows service's purpose be clear.

It's also not really a generic KinesisProxy in the same way that SocketProxy agnostically proxies TCP sockets. It at the very least looks at the stream to parse at EOT and sends a Kinesis message called `ocs_message`. I thought about actually making it a generic KinesisProxy (and open sourced and usable by all), and having the specific OCS behavior swappable by the client, but that seems a bit overkill for our purposes here. Plus, we're wanting this app over time to get _more_ OCS specific and start parsing the messages and uploading different types.

paulswartz: Those are all good points! I've renamed it to `Tricycle` for now, and mentioned that the exact name can be resolved outside the RFC :)

rfcs/0005-kinesis-proxy-json.md
```diff
@@ -0,0 +1,321 @@
...
+- insecure: anyone from the MBTA network can forge OCS messages
+- unreliable: if an RTR instance isn't running (crashed, or during a
+  restart/redeploy) when a message comes in, that message is lost
+- unreliable: if the production RTR instance isn't running, then the log of
+  messages written to S3 (and used by OPMI) will not have those messages
+- wasteful: we send each message four separate times to four RTR instances
+- missing metadata: the messages do not include the date they were sent, or
+  the time the messages were received by SocketProxy 
+
+To address these issues, we'll create Tricycle, a new Elixir application,
```
losvedir: 😂

losvedir:
This looks good to me. This RFC is now in Final Comment Period with a disposition to accept. Since there aren't any outstanding disagreements, I don't think it needs to be long. I will plan to merge tomorrow.

losvedir:
Okay, I'm ready to merge! Can you push up a commit changing the status to "accepted" and putting it in the accepted folder?

PR #6, opened 2021-06-04
RFC 0006 - GTFS canonical routes

losvedir:
This RFC is to propose and discuss the idea of "canonical routes". The motivation for the question is how we should treat routes that have long-term closures, e.g. Brookline Hills on Green-D right now, Lechmere and Science Park, and a some Commuter Rail routes.

Are those stops canonically "part of" the route, even if there is not any scheduled service to them for the foreseeable future? Should clients display those stops even without any predictions? Does it help with alerts?

### DIFF
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -55,9 +55,7 @@ The canonical service appears in the following GTFS files in the following way:
   the canonical trips (added in the trips file below) do not resemble any scheduled service. New
   patterns in this file shall have a new `route_pattern_typicality` value of should be `5`, and
   `route_pattern_sort_order` should be a greater than the value for all non-canonical route patterns
-  for that given route/direction/branch. This will be done by adding two new digits to
-  `route_pattern_sort_order` (one for branch and the other for canonicalness, both placed right
-  after the existing digits corresponding to the `route_sort_order`). For routes with branches,
+  for that given route/direction/branch. For routes with branches,
   you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
   should have _two_ new route patterns in the outbound direction). Finally, a new
   `canonical_route_pattern` field is added to all route patterns, defined with the following values:
@@ -77,18 +75,15 @@ The canonical service appears in the following GTFS files in the following way:
   route patterns.
 - **stop_times.txt** includes all stops that trains in the route/direction passes through, even if
   they are temporarily closed. Increment the arrival/departure times by one minute for each
-  subsequent stop. For a sample, see the attached GTFS file, the output of [✓ 🧪 🚝 Canonical
+  subsequent stop. For a sample, see the GTFS file attached to [✓ 🧪 🚝 Canonical
   trips/patterns experiment](https://app.asana.com/0/881264583703207/1200210504369250), which
   trialled this for Green Line D and Lowell Line (only in the outbound direction).
 
-This will be done as a new "step" of the Makefile called `canonical`, which runs between `truncate`
-and `current`. There will be a new a directory, `input/canonical/` to store any static data the step
-uses.
-
-The canonical step will be a Python module that reads in the `truncate` feed, and writes out its own
-feed after transforming it. The transformations are as above, and also will update any canonical
-trips to use current route patterns (and vice versa) if relevant, eliminating redundant route
-patterns.
+There will be a new a directory, `input/canonical/` to store any static, canonical data that may 
+become published. The determination of what canonical information is to be published will be added 
+as part of the existing `truncate` phase (or made into a new, subsequent `canonical` phase if need 
+be). This step will update any canonical trips to use current route patterns (and vice versa) if 
+relevant, eliminating redundant route patterns.
 
 For Commuter Rail route patterns, the canonical one will specify every stop as non-flag, non-leave
 early stop. For both subway and Commuter Rail canonical trips, no data around bicycle allowances
@@ -100,7 +95,7 @@ This is more complicated than not doing it.
 
 # Rationale and alternatives
 
-See proposed Concepts [1](https://github.com/mbta/technology-docs/pull/6#issuecomment-952315958), [4](https://github.com/mbta/technology-docs/pull/6#issuecomment-952315958), and [5](https://github.com/mbta/technology-docs/pull/6#issuecomment-962120434) in prior discussion.
+See previously-proposed Concepts [1](https://github.com/mbta/technology-docs/pull/6#issuecomment-952315958), [4](https://github.com/mbta/technology-docs/pull/6#issuecomment-952315958), and [5](https://github.com/mbta/technology-docs/pull/6#issuecomment-962120434) in the pull request discussion.
 
 # Prior art
 
@@ -109,13 +104,7 @@ Until now we've been doing ad-hoc adjustments to determine the route pattern of 
 
 # Unresolved questions
 
-Ensuring that we would not be in a situation where the sort order, perhaps ordered by
-frequency/proportion of trips, becomes a different order than desired for the main route pattern
-against the branch route pattern.
-
-How route patterns containing canonical representative trips are tagged or otherwise identified:
-Using a new `canonical_route_pattern` field in route_patterns.txt, or is the new
-`service_schedule_typicality` value sufficient?
+Earlier questions around sorting route patterns, and identifying canonical representative trip, have been resolved.
 
 # Future possibilities
 
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -55,13 +55,14 @@ The canonical service appears in the following GTFS files in the following way:
   the canonical trips (added in the trips file below) do not resemble any scheduled service. New
   patterns in this file shall have a new `route_pattern_typicality` value of should be `6`, and
   `route_pattern_sort_order` should be a greater than the value for all non-canonical route patterns
-  for that given route/direction. This will be done by adding a new digit to
-  `route_pattern_sort_order` (placed right after the existing digits corresponding to the
-  `route_sort_order`). For routes with branches, you'll need a pattern for each
-  route/direction/branch (eg, for Providence/Stoughton Line, you should have _two_ new route
-  patterns in the outbound direction). Finallly, a new `canonical_route_pattern` field is added to
-  all route patterns, indicating whether or not it can be considered canonical (or if the route does
-  not have canonical patterns defined, as will be the case for bus initially).
+  for that given route/direction/branch. This will be done by adding two new digits to
+  `route_pattern_sort_order` (one for branch and the other for canonicalness, both placed right
+  after the existing digits corresponding to the `route_sort_order`). For routes with branches,
+  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
+  should have _two_ new route patterns in the outbound direction). Finally, a new
+  `canonical_route_pattern` field is added to all route patterns, indicating whether or not it can
+  be considered canonical (or if the route does not have canonical patterns defined, as will be the
+  case for bus initially).
 - **shapes.txt** will add new shapes for all canonical subway trips, since the shape IDs can change
   between ratings, especially when stations are open and closed on a long-term basis. These new
   shapes can have an identical set of points as those shapes used during current service. Existing
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -7,8 +7,11 @@
 # Summary
 
 This will enable us to publish in our static GTFS feed "canonical" data for our rail routes. We will
-do so via a "canonical" service that's never active, but which has representative trips, stop times,
-shapes, and route patterns.
+do so via a "canonical" service that's never active, but which has representative trips with stop
+times and shapes. If a canonical trip is similar enough to a real-world trip that are otherwise
+in the GTFS feed, then it will be made the representative trip of an existing route pattern. If no
+such trips are being run (such as due to a long-running station closure), then a new, canonical
+route pattern will be created, of which the canonical trip will be representative.
 
 # Motivation
 
@@ -21,20 +24,20 @@ are able to stay informed about the status of the station closure.
 # Guide-level explanation
 
 GTFS output includes additional, hardcoded data about where rail lines go to and where they can stop
-if all stations were open, which is independent of all scheduled service.
+if all stations were open; these hardcoded data are independent of all scheduled service except that
+they may be tied to scheduled route patterns in some cases.
 
 This is represented as a service with ID "canonical", which is valid for the entire feed date range,
 but which has no active days of the week. The typicality of the service is tagged with a new value
 so it can be programmatically identified.
 
 All rail routes have two or more trips on this service, corresponding to the "canonical" patterns of
-service in each direction for any branches, as understood by stakeholders. Each such trip has a
-route pattern, for which it is the representative trip, and a shape, and the route pattern is
-considered "typical" for the route. If current service has trips following the same route pattern,
-they will use the canonical route pattern as well.
-
-In short, for consumers of the feed, a route pattern has typicality of `1` if and only if it visits
-a set of canonical stops in a canonical order.
+service in each direction for any branches, as understood by stakeholders. Each such canonical trip
+has a route pattern, for which it is the representative trip, and a shape. If current service has
+trips following the same set and order of stops as the canonical trip, then the canonical trip will
+reference—and become the representative trip for—the current service's route pattern. If current
+service has no trips resembling the canonical trip, then a separate, canonical route pattern will be
+maintained in the GTFS output.
 
 # Reference-level explanation
 
@@ -48,15 +51,27 @@ The canonical service appears in the following GTFS files in the following way:
   dates. `service_schedule_typicality` should be `6`, and the
   [docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md) should be updated
   with this new kind of value.
-- **route_patterns.txt** includes new route patterns for all routes/directions we want to hardcode.
-  `route_pattern_typicality` should be `1`, and `route_pattern_sort_order` should be a less than the
-  value for all other route patterns for that given route/direction. For routes with branches,
-  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
-  should have two new route patterns in the outbound direction).
-- **shapes.txt** can probably use the existing shapes.
-- **trips.txt** includes one new trip for each new route pattern. It should reference aforementioned
-  route pattern and shapes. It should reference the new `service_id` created above. Each of these
-  trips should be the `representative_trip_id` for their respective route patterns.
+- **route_patterns.txt** includes new route patterns for routes/directions we want to hardcode when
+  the canonical trips (added in the trips file below) do not resemble any scheduled service. New
+  patterns in this file shall have a new `route_pattern_typicality` value of should be `6`, and
+  `route_pattern_sort_order` should be a greater than the value for all non-canonical route patterns
+  for that given route/direction. This will be done by adding a new digit to
+  `route_pattern_sort_order` (placed right after the existing digits corresponding to the
+  `route_sort_order`). For routes with branches, you'll need a pattern for each
+  route/direction/branch (eg, for Providence/Stoughton Line, you should have _two_ new route
+  patterns in the outbound direction). Finallly, a new `canonical_route_pattern` field is added to
+  all route patterns, indicating whether or not it can be considered canonical (or if the route does
+  not have canonical patterns defined, as will be the case for bus initially).
+- **shapes.txt** will add new shapes for all canonical subway trips, since the shape IDs can change
+  between ratings, especially when stations are open and closed on a long-term basis. These new
+  shapes can have an identical set of points as those shapes used during current service. Existing
+  Commuter Rail shape references can be reused for canonical trips.
+- **trips.txt** includes one new trip for each canonical route/direction pair, or multiple trips in
+  the event of regular branching occuring along the route. It should reference the new `service_id`
+  created above. It should reference either a new (canonical) or previously-existing (regular
+  service) route pattern, depending on whether the trip's stopping pattern is currently served by
+  regular service. Each of these trips should be the `representative_trip_id` for their referenced
+  route patterns.
 - **stop_times.txt** includes all stops that trains in the route/direction passes through, even if
   they are temporarily closed. Increment the arrival/departure times by one minute for each
   subsequent stop. For a sample, see the attached GTFS file, the output of [✓ 🧪 🚝 Canonical
@@ -68,25 +83,21 @@ and `current`. There will be a new a directory, `input/canonical/` to store any 
 uses.
 
 The canonical step will be a Python module that reads in the `truncate` feed, and writes out its own
-feed after transforming it. The transformations are as above, and also will update any current trips
-to use canonical route patterns if relevant, eliminating redundant route patterns.
+feed after transforming it. The transformations are as above, and also will update any canonical
+trips to use current route patterns (and vice versa) if relevant, eliminating redundant route
+patterns.
 
 For Commuter Rail route patterns, the canonical one will specify every stop as non-flag, non-leave
-early stop.
+early stop. For both subway and Commuter Rail canonical trips, no data around bicycle allowances
+will be provided (`bikes_allowed=0`).
 
 # Drawbacks
 
-This is more complicated than not doing it. Some apps may prefer to treat the "typical" route
-pattern as the one that has current service. For example, in the case of a long-term stop closure,
-perhaps some apps don't want to show the closed stop at all.
+This is more complicated than not doing it.
 
 # Rationale and alternatives
 
-We could leave the current route patterns and trips alone, and add canonical route patterns with a
-new typicality value of `5`. Their only trips would be their single representative trip. With this
-approach, since there's no interaction with the remaining data at all, we would make the `canonical`
-feed one that goes into the `merged` step. (Though we'd have to figure out how to determine the
-valid feed start and end date.)
+See proposed Concepts [1](https://github.com/mbta/technology-docs/pull/6#issuecomment-952315958), [4](https://github.com/mbta/technology-docs/pull/6#issuecomment-952315958), and [5](https://github.com/mbta/technology-docs/pull/6#issuecomment-962120434) in prior discussion.
 
 # Prior art
 
@@ -95,8 +106,13 @@ Until now we've been doing ad-hoc adjustments to determine the route pattern of 
 
 # Unresolved questions
 
-The main question is whether current service should be put on canonical route patterns, if
-applicable, or not.
+Ensuring that we would not be in a situation where the sort order, perhaps ordered by
+frequency/proportion of trips, becomes a different order than desired for the main route pattern
+against the branch route pattern.
+
+How route patterns containing canonical representative trips are tagged or otherwise identified:
+Using a new `canonical_route_pattern` field in route_patterns.txt, or is the new
+`service_schedule_typicality` field sufficient?
 
 # Future possibilities
 
@@ -105,4 +121,4 @@ pattern is simply that with a typicality of `1`. In the alternative approach, th
 be `5`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
 its typicality, most relevantly when included with `trips`.
 
-Shoudl we do something similar for ferry and/or bus?
+Should we do something similar for ferry and/or bus?
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -1,7 +1,7 @@
 - Feature Name: gtfs_canonical_routes
 - Start Date: 2021-06-03
 - RFC PR: [mbta/technology-docs#0006](https://github.com/mbta/technology-docs/pull/0006)
-- Asana task: [asana link](https://app.asana.com/)
+- Asana task: [🚝 Publish canonical trips/patterns for rail routes](https://app.asana.com/0/881264583703207/1200325279789524/f)
 - Status: Proposed
 
 # Summary
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -1,6 +1,6 @@
 - Feature Name: gtfs_canonical_routes
 - Start Date: 2021-06-03
-- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
+- RFC PR: [mbta/technology-docs#0006](https://github.com/mbta/technology-docs/pull/0006)
 - Asana task: [asana link](https://app.asana.com/)
 - Status: Proposed
 
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -53,16 +53,18 @@ The canonical service appears in the following GTFS files in the following way:
   with this new kind of value.
 - **route_patterns.txt** includes new route patterns for routes/directions we want to hardcode when
   the canonical trips (added in the trips file below) do not resemble any scheduled service. New
-  patterns in this file shall have a new `route_pattern_typicality` value of should be `6`, and
+  patterns in this file shall have a new `route_pattern_typicality` value of should be `5`, and
   `route_pattern_sort_order` should be a greater than the value for all non-canonical route patterns
   for that given route/direction/branch. This will be done by adding two new digits to
   `route_pattern_sort_order` (one for branch and the other for canonicalness, both placed right
   after the existing digits corresponding to the `route_sort_order`). For routes with branches,
   you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
   should have _two_ new route patterns in the outbound direction). Finally, a new
-  `canonical_route_pattern` field is added to all route patterns, indicating whether or not it can
-  be considered canonical (or if the route does not have canonical patterns defined, as will be the
-  case for bus initially).
+  `canonical_route_pattern` field is added to all route patterns, defined with the following values:
+    - `0` or empty: Route does not have any canonical patterns defined.
+    - `1`: Route pattern should be considered canonical for this route in this direction. If
+      branching regularly occurs, this route-direction may have more than one canonical pattern.
+    - `2`: Route pattern should be not considered canonical for this route in this direction.
 - **shapes.txt** will add new shapes for all canonical subway trips, since the shape IDs can change
   between ratings, especially when stations are open and closed on a long-term basis. These new
   shapes can have an identical set of points as those shapes used during current service. Existing
@@ -117,9 +119,5 @@ Using a new `canonical_route_pattern` field in route_patterns.txt, or is the new
 
 # Future possibilities
 
-We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
-pattern is simply that with a typicality of `1`. In the alternative approach, the typicality would
-be `5`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
-its typicality, most relevantly when included with `trips`.
-
-Should we do something similar for ferry and/or bus?
+We could expose the new data more clearly in the API. For instance, the `service` can't be filtered by
+its typicality, most relevantly when included with `trips`. We could add canonical trips and patterns for ferry and/or bus routes.
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -33,7 +33,7 @@ route pattern, for which it is the representative trip, and a shape, and the rou
 considered "typical" for the route. If current service has trips following the same route pattern,
 they will use the canonical route pattern as well.
 
-In short, for consumers of the feed, a route pattern has typicality of `0` if and only if it visits
+In short, for consumers of the feed, a route pattern has typicality of `1` if and only if it visits
 a set of canonical stops in a canonical order.
 
 # Reference-level explanation
@@ -101,8 +101,8 @@ applicable, or not.
 # Future possibilities
 
 We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
-pattern is simply that with a typicality of `0`. In the alternative approach, the typicality would
-be `4`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
+pattern is simply that with a typicality of `1`. In the alternative approach, the typicality would
+be `5`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
 its typicality, most relevantly when included with `trips`.
 
 Shoudl we do something similar for ferry and/or bus?
--- a/rfcs/0006-gtfs-canonical-routes.md
+++ b/rfcs/0006-gtfs-canonical-routes.md
@@ -113,7 +113,7 @@ against the branch route pattern.
 
 How route patterns containing canonical representative trips are tagged or otherwise identified:
 Using a new `canonical_route_pattern` field in route_patterns.txt, or is the new
-`service_schedule_typicality` field sufficient?
+`service_schedule_typicality` value sufficient?
 
 # Future possibilities
 
--- a/rfcs/0000-gtfs-canonical-routes.md
+++ b/rfcs/0000-gtfs-canonical-routes.md
@@ -1,0 +1,108 @@
+- Feature Name: gtfs_canonical_routes
+- Start Date: 2021-06-03
+- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
+- Asana task: [asana link](https://app.asana.com/)
+- Status: Proposed
+
+# Summary
+
+This will enable us to publish in our static GTFS feed "canonical" data for our rail routes. We will
+do so via a "canonical" service that's never active, but which has representative trips, stop times,
+shapes, and route patterns.
+
+# Motivation
+
+MBTA.com users are able to see line diagrams that match real-world state, including physical maps,
+even when stations are closed for a long duration.
+
+PIOs are able to create alerts for stations without any scheduled service, and T-Alerts subscribers
+are able to stay informed about the status of the station closure.
+
+# Guide-level explanation
+
+GTFS output includes additional, hardcoded data about where rail lines go to and where they can stop
+if all stations were open, which is independent of all scheduled service.
+
+This is represented as a service with ID "canonical", which is valid for the entire feed date range,
+but which has no active days of the week. The typicality of the service is tagged with a new value
+so it can be programmatically identified.
+
+All rail routes have two or more trips on this service, corresponding to the "canonical" patterns of
+service in each direction for any branches, as understood by stakeholders. Each such trip has a
+route pattern, for which it is the representative trip, and a shape, and the route pattern is
+considered "typical" for the route. If current service has trips following the same route pattern,
+they will use the canonical route pattern as well.
+
+In short, for consumers of the feed, a route pattern has typicality of `0` if and only if it visits
+a set of canonical stops in a canonical order.
+
+# Reference-level explanation
+
+The canonical service appears in the following GTFS files in the following way:
+
+- **calendar.txt** includes a new `service_id` to envelop all of the canonical trips on all the
+  routes. `start_date` and `end_date` should reflect the feed's valid start/end dates. All other
+  fields should have a value of `0`.
+- **calendar_attributes.txt** includes the new `service_id`. `service_schedule_type` should be
+  `"Other"`, and `rating_start_date` and `rating_end_date` should reflect the feed's valid start/end
+  dates. `service_schedule_typicality` should be `6`, and the
+  [docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md) should be updated
+  with this new kind of value.
+- **route_patterns.txt** includes new route patterns for all routes/directions we want to hardcode.
+  `route_pattern_typicality` should be `1`, and `route_pattern_sort_order` should be a less than the
+  value for all other route patterns for that given route/direction. For routes with branches,
+  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
+  should have two new route patterns in the outbound direction).
+- **shapes.txt** can probably use the existing shapes.
+- **trips.txt** includes one new trip for each new route pattern. It should reference aforementioned
+  route pattern and shapes. It should reference the new `service_id` created above. Each of these
+  trips should be the `representative_trip_id` for their respective route patterns.
+- **stop_times.txt** includes all stops that trains in the route/direction passes through, even if
+  they are temporarily closed. Increment the arrival/departure times by one minute for each
+  subsequent stop. For a sample, see the attached GTFS file, the output of [✓ 🧪 🚝 Canonical
+  trips/patterns experiment](https://app.asana.com/0/881264583703207/1200210504369250), which
+  trialled this for Green Line D and Lowell Line (only in the outbound direction).
+
+This will be done as a new "step" of the Makefile called `canonical`, which runs between `truncate`
+and `current`. There will be a new a directory, `input/canonical/` to store any static data the step
+uses.
+
+The canonical step will be a Python module that reads in the `truncate` feed, and writes out its own
+feed after transforming it. The transformations are as above, and also will update any current trips
+to use canonical route patterns if relevant, eliminating redundant route patterns.
+
+For Commuter Rail route patterns, the canonical one will specify every stop as non-flag, non-leave
+early stop.
+
+# Drawbacks
+
+This is more complicated than not doing it. Some apps may prefer to treat the "typical" route
+pattern as the one that has current service. For example, in the case of a long-term stop closure,
+perhaps some apps don't want to show the closed stop at all.
+
+# Rationale and alternatives
+
+We could leave the current route patterns and trips alone, and add canonical route patterns with a
+new typicality value of `5`. Their only trips would be their single representative trip. With this
+approach, since there's no interaction with the remaining data at all, we would make the `canonical`
+feed one that goes into the `merged` step. (Though we'd have to figure out how to determine the
+valid feed start and end date.)
+
+# Prior art
+
+Until now we've been doing ad-hoc adjustments to determine the route pattern of routes
+[here](https://github.com/mbta/api/blob/master/apps/state/config/config.exs#L145).
+
+# Unresolved questions
+
+The main question is whether current service should be put on canonical route patterns, if
+applicable, or not.
+
+# Future possibilities
+
+We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
+pattern is simply that with a typicality of `0`. In the alternative approach, the typicality would
+be `4`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
+its typicality, most relevantly when included with `trips`.
+
+Shoudl we do something similar for ferry and/or bus?
--- a/rfcs/accepted/0006-gtfs-canonical-routes.md
+++ b/rfcs/accepted/0006-gtfs-canonical-routes.md
@@ -2,7 +2,7 @@
 - Start Date: 2021-06-03
 - RFC PR: [mbta/technology-docs#0006](https://github.com/mbta/technology-docs/pull/0006)
 - Asana task: [🚝 Publish canonical trips/patterns for rail routes](https://app.asana.com/0/881264583703207/1200325279789524/f)
-- Status: Proposed
+- Status: Accepted
 
 # Summary
 


rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+but which has no active days of the week. The typicality of the service is tagged with a new value
+so it can be programmatically identified.
+
+All rail routes have two or more trips on this service, corresponding to the "canonical" patterns of
+service in each direction for any branches, as understood by stakeholders. Each such trip has a
+route pattern, for which it is the representative trip, and a shape, and the route pattern is
+considered "typical" for the route. If current service has trips following the same route pattern,
+they will use the canonical route pattern as well.
+
+In short, for consumers of the feed, a route pattern has typicality of `0` if and only if it visits
```
digitalcora: I think you meant `1` here? It says `1` in the reference explanation below, but `0` is also a valid value (though it just means "undefined").

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+
+MBTA.com users are able to see line diagrams that match real-world state, including physical maps,
+even when stations are closed for a long duration.
+
+PIOs are able to create alerts for stations without any scheduled service, and T-Alerts subscribers
+are able to stay informed about the status of the station closure.
+
+# Guide-level explanation
+
+GTFS output includes additional, hardcoded data about where rail lines go to and where they can stop
```
digitalcora: minor tense thing: "where they _would_ stop if all stations were open"

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+For Commuter Rail route patterns, the canonical one will specify every stop as non-flag, non-leave
+early stop.
+
+# Drawbacks
+
+This is more complicated than not doing it. Some apps may prefer to treat the "typical" route
+pattern as the one that has current service. For example, in the case of a long-term stop closure,
+perhaps some apps don't want to show the closed stop at all.
+
+# Rationale and alternatives
```
digitalcora: This section seems to be missing the "rationale" part.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+
+We could leave the current route patterns and trips alone, and add canonical route patterns with a
+new typicality value of `5`. Their only trips would be their single representative trip. With this
+approach, since there's no interaction with the remaining data at all, we would make the `canonical`
+feed one that goes into the `merged` step. (Though we'd have to figure out how to determine the
+valid feed start and end date.)
```
digitalcora: Could you explain the benefits and drawbacks here compared to the "chosen" approach?

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+  fields should have a value of `0`.
+- **calendar_attributes.txt** includes the new `service_id`. `service_schedule_type` should be
+  `"Other"`, and `rating_start_date` and `rating_end_date` should reflect the feed's valid start/end
+  dates. `service_schedule_typicality` should be `6`, and the
+  [docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md) should be updated
+  with this new kind of value.
+- **route_patterns.txt** includes new route patterns for all routes/directions we want to hardcode.
+  `route_pattern_typicality` should be `1`, and `route_pattern_sort_order` should be a less than the
+  value for all other route patterns for that given route/direction. For routes with branches,
+  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
```
paulswartz: Should we define the sort order for branches? (my vote would be in the order of the names, so the Providence branch would have a lower sort order, or alphabetical, so Alewife would be lower than Braintree)

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+This will be done as a new "step" of the Makefile called `canonical`, which runs between `truncate`
+and `current`. There will be a new a directory, `input/canonical/` to store any static data the step
+uses.
+
+The canonical step will be a Python module that reads in the `truncate` feed, and writes out its own
+feed after transforming it. The transformations are as above, and also will update any current trips
+to use canonical route patterns if relevant, eliminating redundant route patterns.
+
+For Commuter Rail route patterns, the canonical one will specify every stop as non-flag, non-leave
+early stop.
```
paulswartz: Hmm. We have some CR routes where all trips at a stop are flag stops. For example, Ruggles and Back Bay on Needham: https://www.mbta.com/schedules/CR-Needham/timetable

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+applicable, or not.
+
+# Future possibilities
+
+We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
+pattern is simply that with a typicality of `0`. In the alternative approach, the typicality would
+be `4`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
+its typicality, most relevantly when included with `trips`.
+
+Shoudl we do something similar for ferry and/or bus?
```
paulswartz: should

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+We could leave the current route patterns and trips alone, and add canonical route patterns with a
+new typicality value of `5`. Their only trips would be their single representative trip. With this
+approach, since there's no interaction with the remaining data at all, we would make the `canonical`
+feed one that goes into the `merged` step. (Though we'd have to figure out how to determine the
+valid feed start and end date.)
+
+# Prior art
+
+Until now we've been doing ad-hoc adjustments to determine the route pattern of routes
+[here](https://github.com/mbta/api/blob/master/apps/state/config/config.exs#L145).
```
digitalcora: I don't think this tells the whole story — this is configuration in the API, so it doesn't explain the GTFS side, and it's not directly declaring a "canonical" route pattern for routes.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+  trips should be the `representative_trip_id` for their respective route patterns.
+- **stop_times.txt** includes all stops that trains in the route/direction passes through, even if
+  they are temporarily closed. Increment the arrival/departure times by one minute for each
+  subsequent stop. For a sample, see the attached GTFS file, the output of [✓ 🧪 🚝 Canonical
+  trips/patterns experiment](https://app.asana.com/0/881264583703207/1200210504369250), which
+  trialled this for Green Line D and Lowell Line (only in the outbound direction).
+
+This will be done as a new "step" of the Makefile called `canonical`, which runs between `truncate`
+and `current`. There will be a new a directory, `input/canonical/` to store any static data the step
+uses.
```
digitalcora: I'd be interested to hear (perhaps in the "Rationale") why this method would be preferred over introducing the "canonical" schedule early in the process, and having subsequent steps that would normally create route patterns check whether they're duplicating a canonical one.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+
+The canonical step will be a Python module that reads in the `truncate` feed, and writes out its own
+feed after transforming it. The transformations are as above, and also will update any current trips
+to use canonical route patterns if relevant, eliminating redundant route patterns.
```
digitalcora: How will we guard against "drift" where trips change away from the canonical pattern, which could quietly create new patterns when we actually want to update the canonical one? This would be especially relevant if we later expanded this to bus routes, of which there are way too many to manually check.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+- **calendar_attributes.txt** includes the new `service_id`. `service_schedule_type` should be
+  `"Other"`, and `rating_start_date` and `rating_end_date` should reflect the feed's valid start/end
+  dates. `service_schedule_typicality` should be `6`, and the
+  [docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md) should be updated
+  with this new kind of value.
+- **route_patterns.txt** includes new route patterns for all routes/directions we want to hardcode.
+  `route_pattern_typicality` should be `1`, and `route_pattern_sort_order` should be a less than the
+  value for all other route patterns for that given route/direction. For routes with branches,
+  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
+  should have two new route patterns in the outbound direction).
```
digitalcora: What happens if this results in more than one route pattern with typicality `1` per route/direction/branch? (maybe ties into the "drift" question)

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Unresolved questions
+
+The main question is whether current service should be put on canonical route patterns, if
+applicable, or not.
+
+# Future possibilities
+
+We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
+pattern is simply that with a typicality of `0`. In the alternative approach, the typicality would
```
digitalcora: As above I think you meant `1` here?

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+be `4`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
```
digitalcora: The alternative mentions a new value of `5`, not `4`.

losvedir:
Good questions here. This RFC came mostly transcribed from the Asana task, but I think there's basically a key GTFS question here that @jfabi should decide on, since that would influence the implementation.

@jfabi - Should canonical route patterns be a totally separate "platonic" thing, or should they just be the actual typical route patterns which real trips can use?

I think in the former implementation we'd have a single trip per canonical route-pattern, and the route pattern would have maybe a new `route_pattern_typicality` of `5`. Consumers of GTFS would by default probably de-prioritize or filter out these for the most part, since this is a new, large typicality value. But we could add a filter to the API (or we could rely on the client retrieving all route patterns and finding `5`'s themselves), if they were interested in the canonical patterns. E.g., maybe dotcom would request them to generate their line pages.

In the latter implementation, a canonical route pattern would essentially be *defined* as the typical route pattern for rail routes, and vice versa. They would have `route_pattern_typicality` of `1` and we'd put actual trips on these patterns, in addition (or instead of?) the canonical ones, if they matched – for some definition of "matched", CR TBD. In this case, consumers couldn't help but use the canonical route patterns, and we wouldn't need to carve out a separate category for them. We'd be taking the philosophical stand that although all the trips on this route are for now skipping this closed station, we still call them deviations from the regular route. All naively-generated line pages would automatically include long-term closed stations. This sounds correct to me at an intuitive level, since most riders probably think of them as still part of the route, but I'm not sure.

I think that decision could affect implementation, so let's decide that and then we can figure out where in the process we generate this data.

rfcs/0006-gtfs-canonical-routes.md
jfabi: I'd get behind setting the canonical trip to have L-stops (or F-stops) in cases like these where every real trip behaves the same way.

Broadly, we could define what the behavior should be for the canonical trip if some actual trips have different attributes from others. So for flag stops, we could default to not having a flag stop; for `bikes_allowed`, we could default to `0` if the CR ever goes back to prohibiting bikes during peak hours.

rfcs/0006-gtfs-canonical-routes.md
jfabi: I don't actually _think_ we're doing any special `route_pattern_typicality` overrides in GTFS-creator itself, though we do validate that the output looks as we expect it to: https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/validate.py#L450.

Either way, it's also a good reminder to think a little more about how this will be implemented in the API for dotcom and others to use/not use.

jfabi: Actually, we do at least one manual `route_pattern_typicality` override for Commuter Rail: https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/commuter_rail/route_patterns.py#L64.

rfcs/0006-gtfs-canonical-routes.md
jfabi: I was sort of originally assuming (though it's not spelled out here) that any _non-canonical_ route patterns previously having `route_pattern_typicality=1` would have its value changed to `2` during the `canonical` stage. Would this address the concern?

This also might be make is _slightly_ easer to implement the `canonical` step where originally proposed, rather than at the very beginning, since we assign the original typicality values based on proportion of trips operated.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+  `"Other"`, and `rating_start_date` and `rating_end_date` should reflect the feed's valid start/end
+  dates. `service_schedule_typicality` should be `6`, and the
+  [docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md) should be updated
+  with this new kind of value.
+- **route_patterns.txt** includes new route patterns for all routes/directions we want to hardcode.
+  `route_pattern_typicality` should be `1`, and `route_pattern_sort_order` should be a less than the
+  value for all other route patterns for that given route/direction. For routes with branches,
+  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
+  should have two new route patterns in the outbound direction).
+- **shapes.txt** can probably use the existing shapes.
```
jfabi: This won't always work for Subway, since the shapes (and `shape_id`s) can change between ratings. We might want to create our own duplicate shapes and store them in `input/canonical/shapes.txt`.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+- **trips.txt** includes one new trip for each new route pattern. It should reference aforementioned
+  route pattern and shapes. It should reference the new `service_id` created above. Each of these
+  trips should be the `representative_trip_id` for their respective route patterns.
+- **stop_times.txt** includes all stops that trains in the route/direction passes through, even if
```
jfabi: In the same vein as shapes, a long-term subway closure could mean that the closed `stop_id`s could be missing from `build/truncate/stops.txt`, which is why https://github.com/mbta/gtfs_creator/pull/1243 was needed. So we could also have an `input/canonical/stops.txt` that includes the stops...

...But if we go ahead to just manually specify every shape (if we used the same `shape_id`s as revenue trips) and every stop ID in `input/canonical`, we'd now have to drop most of them as duplicates. Would that mean this module is better off coming _right before_ `truncate`, instead of after?

jfabi (commented):
>@jfabi - Should canonical route patterns be a totally separate "platonic" thing, or should they just be the actual typical route patterns which real trips can use?

I'll keep mulling this over the next couple days, but as far as my current thinking:

I generally do prefer the ideal of the Option :two: (latter implementation)—the canonical pattern should just be like any other route pattern, and should have revenue trips if any happen to be running on it.

My hangup on it is the 🚌 question: I don't see an easy course for scaling canonical routes to include bus routes, any would either involve heavy modification to our HASTUS export, or plenty of manual maintenance by the TRC team. Changes to a service on a route between ratings could be due to a 3-month construction project, due to a permanent street layout change, or maybe it's because of [Bus Network Redesign](https://www.mbta.com/projects/bus-network-redesign). This gives `route_pattern_typicality=1` a different meaning between Options 1️⃣ and 2️⃣ (for bus, it'd only "canonical" within the current season).

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,108 @@
...
+applicable, or not.
+
+# Future possibilities
+
+We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
+pattern is simply that with a typicality of `1`. In the alternative approach, the typicality would
+be `5`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
+its typicality, most relevantly when included with `trips`.
+
+Shoudl we do something similar for ferry and/or bus?
```
jfabi: Future v3 idea related to bus (if we don't have canonical trips stored for every bus route): Whenever we add a ["HASTUS patch" for bus](https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/hastus/update.py#L1243), we automatically create a canonical route pattern with the pre-patched stop list.

jfabi:
_Another, more backwards-compatible, alternative I thought of, to chew on, that is a variation of Option 2️⃣. Call it Option 3️⃣:_

A new `route_pattern_typicality=5` pattern _only_ make it into the published feed if it's distinct from the otherwise-typical pattern.

This way, if you're a consumer that _wants_ to show the new "canonical patterns", you first look for `5` patterns, and then if you find none, you revert to look for `1` patterns (or you could simply use `route_pattern_sort_order`).

If you're a consumer that only wants to look at operated service, you continue your current behavior of looking for `1` patterns first. I'm thinking maybe of analysts (eg, in OPMI) who may be comparing service between different ratings—they have asked me in the past about how to use pattern `typicality` and `sort_order`, so I want to assume they would be stakeholders here as well.

I guess yet another option (call it Option 4️⃣) on top of 3️⃣ might be to introduce an independent sort of `canonical_pattern` boolean. I _think_ the only advantage would be to make obtaining the canonical patterns slightly more easily, which I'm not sure is really necessary.

Thoughts?

thecristen:
As an API consumer, having this available through the route_pattern typicality value conceptually makes the most sense to me. Thanks for this proposal!

Am late to the discussion but I like the idea of a novel, service-independent `route_pattern_typicality` value as @jfabi described. I think it makes sense to offer canonical routes as an opt-in, you really have to be wanting it to retrieve it sort of deal. 

I'm not sure how the representative trip value would work when there happens to be no active service that is following the canonical pattern. Sorry if I missed it in the document.

digitalcora:
> I'm not sure how the representative trip value would work when there happens to be no active service that is following the canonical pattern. Sorry if I missed it in the document.

My understanding: In the event we added a new typicality value for canonical patterns, there would only ever be a single "fake" trip on each of those patterns, for the sake of giving them a shape / stop list. The scheduled stop times would not necessarily be meaningful. In other words, it would almost be a dedicated API for getting _just_ the canonical shape/stops for a route, but fit into the existing data structures.

lemald:
One thought that just occurred to me (which is a bit of a long shot, but maybe worth considering briefly) is that we _do_ have a use for a canonical ordering of timepoints on a bus route in Skate for the purposes of drawing route ladders. One immediate issue I see is that the route ladder for a route includes all timepoints, not just those on one particular variant, so there won't necessarily be any one trip that hits all of the timepoints. This is also somewhat different from the rail case, since we're not representing a piece of physical infrastructure. Thoughts? Don't be afraid to say that this is totally unrelated and shouldn't be considered as part of this RFC. 🙂

losvedir:
> One immediate issue I see is that the route ladder for a route includes all timepoints, not just those on one particular variant, so there won't necessarily be any one trip that hits all of the timepoints.

This doesn't sound like too big of an issue to me. A canonical route doesn't actually need a real trip to be on it, in order to be canonical. In fact, that's kind of the purpose of this idea in the first place, to include stations that are closed and that no real trip visits at the moment. We're talking about adding a single pseudo-trip with 1 minute transit times between stops per canonical route to reify it.

And actually, I guess we're talking about "canonical route *patterns*", since there can be multiple of them per route. For example, we'd have two each direction on the Red line, one for Ashmont and one for Braintree branches. (This does leave some work on the client end to integrate them in a way that works for their use case.)

So I think we could cover your route ladder by some combination of multiple canonical route patterns + pseudo-trips that hit stops on each one.

The idea of how to extend this to bus has been kicked around a bit, and we weren't super sure, especially since it can change each rating. What "is" a bus route even? The existence of Skate's route ladders does point to there being some sort of external intuition about what certain routes are. Can you talk a bit about how you generate them now? Is it entirely algorithmic? Are there inputs from yourselves or users based on what you feel should be on the ladder? Do they change from rating to rating or within ratings?

lemald:
> Can you talk a bit about how you generate them now? Is it entirely algorithmic? Are there inputs from yourselves or users based on what you feel should be on the ladder? Do they change from rating to rating or within ratings?

Good questions! We have an algorithm that works most of the time, and an additional file where we can define overrides. Usually we only add these overrides when we get user feedback about the order being "wrong," but it's not always clear what the canonical source of that order is - I think sometimes it may just be "wrong" in the sense of not matching TransitMaster, which just has manually-defined timepoint orders for each route. I'll ask around and see how much we know about that.

jfabi:
Based on the above discussion and feedback, I'm going to narrow us down to two primary options:

**Concept 1️⃣ — fully-separate canonical patterns:**
- New `route_pattern_typicality` value of `5` for canonical patterns introduced to `route_patterns.txt`.
  - _Possible definition:_ "Pattern contains all permanent stops known to riders to be on the route, including those which may be temporarily closed."
- New trips, stop times, and route patterns (with the new `5` value) are introduced for all routes needing a canonical set of stops served (ie, canonical pattern). These new trips would not be activated for any dates/days of the week.
- _Pros?:_ Keeps canonical route patterns, and those being used for scheduled service, fully separated, making it fool-proof easy to identify the canonical patterns for a given route, if they exist.
- Sample `route_patterns.txt` for the Green Line in one direction:
```
route_id,direction_id,route_pattern_id,route_pattern_typicality,route_pattern_name
Green-B,0,Green-B-812-0,1,Government Center - Boston College
Green-B,0,Green-B-canonical-0,5,Government Center - Boston College
Green-C,0,Green-C-832-0,1,Government Center - Cleveland Circle
Green-C,0,Green-C-canonical-0,5,Government Center - Cleveland Circle
Green-D,0,Green-D-851-0,1,North Station - Riverside (omit Brookline Hills)
Green-D,0,Green-D-841-0,3,North Station - Reservoir (omit Brookline Hills)
Green-D,0,Green-D-canonical-0,5,North Station - Riverside
Green-E,0,Green-D-885-0,1,North Station - Heath Street
Green-E,0,Green-E-canonical-0,5,Union Square - Riverside
```

**Concept 4️⃣  — separate canonical patterns if otherwise missing:**
- New `route_pattern_typicality` value of `5` for canonical patterns introduced to `route_patterns.txt` because there is no existing pattern in service matching the canonical set of stops that should be on the route.
  - _Possible definition:_ "Pattern is not currently operated, but may contain temporarily-closed stops."
  - Note that we could just as easily use `route_pattern_typicality=0` under this concept.
- New trips, stop times, and route patterns are introduced for any rail routes not already having their canonical pattern in the service. These new trips would not be activated for any dates/days of the week.
- Introduce a new `canonical_route_pattern` to make it crystal clear whether a "typical" pattern is also the "canonical" one.
  - This may especially handy if we have modes that do not explicitly have identified canonical patterns (eg, bus), or if there are routes which have canonical patterns yet lack `route_pattern_typicality=1` patterns.
  - It could also head off any ambiguity around how to treat routes with multiple branches (eg, Red Line).
- _Pros?:_ If a canonical pattern is being actively used in service, this can be easily identified.
- Sample `route_patterns.txt` for the Green Line in one direction:
```
route_id,direction_id,route_pattern_id,route_pattern_typicality,route_pattern_name,canonical_route_pattern
Green-B,0,Green-B-812-0,1,Government Center - Boston College,1
Green-C,0,Green-C-832-0,1,Government Center - Cleveland Circle,1
Green-D,0,Green-D-851-0,1,North Station - Riverside (omit Brookline Hills),
Green-D,0,Green-D-841-0,3,North Station - Reservoir (omit Brookline Hills),
Green-D,0,Green-D-canonical-0,5,North Station - Riverside,1
Green-E,0,Green-D-885-0,1,North Station - Heath Street,
Green-E,0,Green-E-canonical-0,5,Union Square - Riverside,1
```

**Sample status quo `route_patterns.txt`, for comparison:**
```
route_id,direction_id,route_pattern_id,route_pattern_typicality,route_pattern_name
Green-B,0,Green-B-812-0,1,Government Center - Boston College
Green-C,0,Green-C-832-0,1,Government Center - Cleveland Circle
Green-D,0,Green-D-851-0,1,North Station - Riverside (omit Brookline Hills)
Green-D,0,Green-D-841-0,3,North Station - Reservoir (omit Brookline Hills)
Green-E,0,Green-D-885-0,1,North Station - Heath Street
```

jfabi:
Welcome any suggestions and other feedback that folks—both those in the TRC sphere and potential consumers of these data—may have about these two concepts (or others)! 🎉

digitalcora:
Concept 4️⃣ sounds best to me, since it allows separating "is this pattern considered canonical" from "how typical is this pattern in scheduled service". It's then easy to determine, for example, when a route's "canonical" pattern is not one that has any scheduled service. To figure this out under concept 1️⃣ you'd have to look at all patterns for the route, checking whether there are non-`5` patterns with the same stops and shapes as `5` patterns.

Within concept 4️⃣, I think it makes sense to use `5` as the new typicality value, since it follows the pattern of higher typicalities being less common in scheduled service (in this concept, the only time such a pattern would exist is when it has _no_ scheduled service at all, otherwise we would have applied the "canonical" flag to one of the existing patterns that does have service).

paulswartz:
@jfabi 
> **Concept 4️⃣ — separate canonical patterns if otherwise missing:**
>     * Introduce a new `canonical_route_pattern` to make it crystal clear whether a "typical" pattern is also the "canonical" one.
>       * This may especially handy if we have modes that do not explicitly have identified canonical patterns (eg, bus), or if there are routes which have canonical patterns yet lack `route_pattern_typicality=1` patterns.

Would the value for bus be `1`, or not provided?

> * It could also head off any ambiguity around how to treat routes with multiple branches (eg, Red Line).

How does this approach treat routes with multiple branches? For example, what would it have looked like when Wollaston was closed?

losvedir:
My gut preference is 1️⃣ , since that seems simpler to implement... I guess I'd like to step back and ask again how this will be used, and who will benefit? I don't want to spend a lot of time on this feature if no one uses it. I found a few asana tasks in the past that mentioned how this could have helped:

* [Broken ferry schedules](https://app.asana.com/0/search/1201313085292316/1200524427366020)
* [Adding alerts to closed stops](https://app.asana.com/0/search/1201313085292316/1180111330864155)
* [Truncating out shapes without trips](https://app.asana.com/0/search/1201313085292316/1131797534112041)
* [API stops-on-route after new Green Line terminals](https://app.asana.com/0/584764604969369/1201281007866196)

I'd love if people could share more examples.

But, as far as I can tell, I don't think there's much benefit in trying to map actual service onto canonical routes, or making that particularly easy for clients. Rather, you'd reach for canonical route patterns when you want to enumerate what the stops are on a route, for example on a line diagram, or for alerts-ui to produce a list of stops for a given selected route.

jfabi:
@paulswartz Considering Concept :four:...

> > * Introduce a new `canonical_route_pattern` to make it crystal clear whether a "typical" pattern is also the "canonical" one.
> > * This may especially handy if we have modes that do not explicitly have identified canonical patterns (eg, bus), or if there are routes which have canonical patterns yet lack `route_pattern_typicality=1` patterns.
> 
> Would the value for bus be `1`, or not provided?

The value would either be not provided, or we could create differentiated `0` and `2` values, with one of the values meaning that this route pattern is explicitly _non_ canonical, and the other one meaning that the route does not have any defined canonical patterns (ie, it's "ambiguous").

> > * It could also head off any ambiguity around how to treat routes with multiple branches (eg, Red Line).
> 
> How does this approach treat routes with multiple branches? For example, what would it have looked like when Wollaston was closed?

It would look similar to the Green Line example, but just changing the `route_id`s to all be the same. So, like this:
```
route_id,direction_id,route_pattern_id,route_pattern_typicality,route_pattern_name,canonical_route_pattern
Red,0,Red-A-0,1,Alewife - Ashmont,1
Red,0,Red-B-0-70198_70200,1,Alewife - Braintree (omit Wollaston),
Red,0,Red-canonicalB-0,5,Alewife - Braintree,1
```

jfabi:
@losvedir

>But, as far as I can tell, I don't think there's much benefit in trying to map actual service onto canonical routes, or making that particularly easy for clients. Rather, you'd reach for canonical route patterns when you want to enumerate what the stops are on a route, for example on a line diagram, or for alerts-ui to produce a list of stops for a given selected route.

Great point! I originally thought that maybe it'd be easier to pull patterns for the line diagrams/maps (which show real-time vehicles), though @thecristen would certainly be a far better expert here!

Also, tagging @AlissaZimmer for any possible feedback and whether there are any specific OPMI use cases that would be helpful to catalog here.

losvedir:
Another use case from Sam Hickey at IBI, from a call just now:

* Order of bus stops in Alerts-UI when a bus route has many variants.

He said they'd be open to using one of our GTFS extensions to solve this issue.

One thing this brings up is if we need to worry about how to combine multiple canonical route patterns. For example, we'd certainly have a canonical route pattern for each of the Red Line branches Ashmont and Braintree. Do we need to (meta-?) canonically determine how they should be combined for the "Red Line" stops? Should we ask a client to infer it based on shared stops and where they diverge? What about the case of Green line where Park St has different stop IDs for each route, there in the middle? (I guess technically those are all different routes, so maybe this isn't an issue for us, but it could be in theory).

And, for completeness, I just realized we should at least discuss a bit, if only to rule out, let's call it **Concept 5️⃣ - New Resource** - i.e. *not* shoehorning this into `route_patterns.txt`. Backtracking a bit - if all we're using this for is something like line pages or alerts UI stop pickers, we could introduce a `stop_order.txt` or something to that effect, and leave off the inactive service and placeholder trip altogether.

losvedir:
Another possible use case: [Line Diagram | CR route "main branch" should be consistent](https://app.asana.com/0/385363666817452/1201138780642048). A bug report from dotcom about a new GTFS rating changing the "primary" (the default un-collapsed on the line page) branch of Commuter Rail lines. Again, relating to displaying a line diagram. Touches on my previous comment of 

>  Do we need to (meta-?) canonically determine how they should be combined for the "Red Line" stops? Should we ask a client to infer it based on shared stops and where they diverge?

and answers that question with a "no". We probably want to have the ability to be explicit in how lines with branches work; if there's a primary or not, and how they diverge and such.

digitalcora:
> We probably want to have the ability to be explicit in how lines with branches work; if there's a primary or not, and how they diverge and such.

It seems to me like route patterns already handle this, between the `typicality` and `sort_order` fields. Yes, you have to determine the graph structure yourself by traversing the (parent) stops on each pattern, but it's not like you're missing any data you would need to do this accurately. I think this is a tangent to the RFC, despite line diagrams being involved in both cases.

> One thing this brings up is if we need to worry about how to combine multiple canonical route patterns. For example, we'd certainly have a canonical route pattern for each of the Red Line branches Ashmont and Braintree. Do we need to (meta-?) canonically determine how they should be combined for the "Red Line" stops?

The API already has a bunch of logic for determining and overriding what "the stops" are for a route and the order they should be listed in, which I assume we would keep using (though some of it could be simplified or deleted) with canonical patterns. Not sure if you meant this in a GTFS context, though I'm not sure if we _should_ be looking to port the stops-on-route concept from the API into GTFS.

thecristen:
> It seems to me like route patterns already handle this, between the `typicality` and `sort_order` fields.

Yeah, kind of, but route patterns are dynamic and it'd be nice to not have to rewrite my graph-constructing logic tomorrow if the route patterns change because of a shuttle replacing part of a branch... hence, the beauty of the canonical route that's not subject to service availability and typicality.

Though I sort of agree that that's tangential... at the very least, I'll note that I'm happy to infer the line diagram structure and piece it together myself, especially if assuming stops-on-route returns stops in an intentional order (stops-on-route + shapes being a foundation from which we create the line diagram currently).

digitalcora:
@thecristen Yes, I think we're in agreement here — my point was just that, on the consumer side, route patterns and their `sort_order` already fully accommodate the need to draw correct line diagrams _other_ than not having "canonical" patterns made available/indicated as per Concepts 1️⃣ and 4️⃣ (interpreting @losvedir's suggestion to _"be explicit in how lines with branches work; if there's a primary or not, and how they diverge and such"_ as something beyond that, but maybe I read this wrong and we're also in agreement). On the producer side we can tweak the `sort_order` to suggest a "main branch" for line diagrams, but this doesn't require anything new for consumers.

> I'll note that I'm happy to infer the line diagram structure and piece it together myself, especially if assuming stops-on-route returns stops in an intentional order (stops-on-route + shapes being a foundation from which we create the line diagram currently).

Stops-on-route does have an intentional order which we've manually overridden in several places, I assume specifically to make the dotcom line diagrams come out right. That said, IMO we should deprecate stops-on-route in favor of canonical route patterns once they're introduced, as they essentially provide the same information (the internal value that determines whether a stop is "on" a route is even called `canonical`), but stops-on-route has a lot of hidden "magic" and is difficult to work with when a route has branches. And, of course, it's not in GTFS.

In the ideal world where we have canonical route patterns and are on the latest API version (so no stops-on-shapes), building a line diagram using the API might look like `GET /route_patterns?filter[route]=Red&filter[direction_id]=0&filter[is_canonical]=true&include=representative_trip.stops.parent_station` and loading all the stops-or-parent-stations into a digraph with edges drawn according to the order of the stops on each pattern — at that point you have a full directional "diagram" of the route, which has all the required information to render a visual diagram. In fact, except for the missing `is_canonical` filter, you can already do this, but you have to guess which patterns are canonical based on either typicality or cross-referencing with the stops-on-route list.

jfabi:
>And, for completeness, I just realized we should at least discuss a bit, if only to rule out, let's call it Concept 5️⃣ - New Resource - i.e. not shoehorning this into route_patterns.txt. Backtracking a bit - if all we're using this for is something like line pages or alerts UI stop pickers, we could introduce a stop_order.txt or something to that effect, and leave off the inactive service and placeholder trip altogether.

I agree that we *could* add this canonical information into a brand new resource, such as a `stop_order.txt`. In fact, we sort of have precedent for doing so: Our special [`stop_order_exceptions.txt`](https://github.com/mbta/gtfs_creator/blob/master/input/alerts_ui/stop_order_exceptions.txt) file in the Alerts UI version that helps sort out some stops for that application.

I'd lean against this approach because, while it most likely would be simpler for us on the TRC side to implement, it would still be a whole new concept for consumers (namely dotcom and Alerts UI) to ingest, delaying the benefits. Dotcom is already set up to read and use route patterns (with stops/shapes relationships being deprecated in the API), and a proof-of-concept has confirmed that Alerts UI can add stops to its route listing even based on unscheduled trips. In Alerts UI, the `stop_order_exceptions.txt` file only helps to sort out stops already on associated with a route, and currently cannot manually _add_ stops to a route.

Such a `stop_order.txt` file would ideally be able to be paired with a shape some way to represent the canonical route's path to/from unserved stops on line maps (see [Green Line E](https://mbta.com/schedules/Green-E) as an example). If the website wanted to show GLX's future routing ahead of opening, for instance, we'd need such a link to shapes.

Regarding the `route_patterns.txt` concepts, I'm happy to pursue either of Concepts :one: or :four:. We don't seem to have established a use case for linking current service to canonical patterns, so I'm happy to drop the Concept :four: if the former is going to be lot simpler for everyone.

Thoughts, @losvedir or others, on a new stop route file and/or next steps? Maybe there's something I'm missing?

digitalcora:
> Regarding the `route_patterns.txt` concepts, I'm happy to pursue either of Concepts 1️⃣ or 4️⃣. We don't seem to have established a use case for linking current service to canonical patterns, so I'm happy to drop the Concept 4️⃣ if the former is going to be lot simpler for everyone.

I'm unsure if concept 1️⃣ is necessarily a _lot_ simpler for everyone. For consumers, the effort involved is about the same (GTFS: either consume a special-case value of typicality, or a new column of route patterns / API: either make a request with `filter[typicality]=5`, or `filter[is_canonical]=true`). For us, the producers, the work of crafting the canonical patterns is identical, with one extra step in concept 4️⃣ where we look at whether a canonical pattern mirrors an existing one, and if so throw it away and set the "is canonical" field on the existing pattern instead.

While it's true we don't yet have a concrete use case for the extra information provided by concept 4️⃣, I think that A) it's likely we'll encounter one in the future, B) back-compat-wise it would be tricky to change from one approach to the other after it's implemented, and C) this approach retains the existing semantics of the `typicality` field, making it easier to consume "generically" (without hard-coding the meaning of each value), and preserving back-compat for any consumers already doing that. For these reasons I feel like concept 4️⃣, while having a slightly higher one-time cost to implement, is the approach least likely to cause problems or require additional work later on.

That's my case — regardless of what you choose, @jfabi, I think the next step should be to revise the actual RFC (anyone remember this was a pull request? 😄) with the new proposed approach, and also address any outstanding comments on the original text, of which there are a few that seem like they'd still be applicable.

losvedir:
Thanks everyone, yep, I think we can rule out my 5️⃣ . The main issue of "main branches" can be solved via the route_pattern's sort order.

Regarding 1️⃣ vs 4️⃣ , I just think the main difference to me is that I think of the "canonical" data as platonic, mostly firewalled from the other data. I don't want "canonical" trips to be tied to the vagaries of calendar activation, or incidentally provide information about bikes, leave-early, or flag stops. I don't feel strongly particularly about how we tag canonical trips, whether it's a new column or a new typicality value.

What about a blend, basically what @digitalcora said, but where the "final step" also updates the representative trip?

* Generate "canonical" route patterns, trips, shapes, stop times, etc.
* Identify "mirrored" route patterns
* Throw away the canonical pattern
* Update the service route pattern to point to the canonical trip (and vice versa)

One conceptual question about tying the canonical-ness to existing route patterns: could we be in a situation where the sort order, perhaps ordered as we do by frequency/proportion of trips, is in a different order to what we want for the main route pattern vs the branch route pattern?

digitalcora:
> Regarding 1️⃣ vs 4️⃣ , I just think the main difference to me is that I think of the "canonical" data as platonic, mostly firewalled from the other data. I don't want "canonical" trips to be tied to the vagaries of calendar activation, or incidentally provide information about bikes, leave-early, or flag stops. I don't feel strongly particularly about how we tag canonical trips, whether it's a new column or a new typicality value.
>
> What about a blend, basically what @digitalcora said, but where the "final step" also updates the representative trip?

I don't completely understand this reasoning given you're referring to "canonical trips" but it's the _pattern_ we're designating "canonical", not any specific trip. And I think it should be understood that not all of a pattern's trips will be exactly the same as far as flag stops, etc.

jfabi:
I made some new commits to this branch reflecting my interpretation of Gabe's blended proposal. @losvedir @digitalcora (and anyone else interested!) Please give the current changes a look and advise if there are any questions, misunderstandings, etc. I also left two discussion points under "Unresolved questions":

>Ensuring that we would not be in a situation where the sort order, perhaps ordered by frequency/proportion of trips, becomes a different order than desired for the main route pattern against the branch route pattern.

I offered a solution to this in the **route_patterns.txt** by expanding the length of `route_pattern_sort_order`, though we would probably also want to specify somewhere in our documentation that the best practice (and indeed the one used by dotcom) is to use the field as a tiebreaker in place of `route_pattern_typicality`.

>How route patterns containing canonical representative trips are tagged or otherwise identified: Using a new `canonical_route_pattern` field in route_patterns.txt, or is the new `service_schedule_typicality` value sufficient?

In this branch's draft I added a `canonical_route_pattern` field to **route_patterns.txt** as a default. @thecristen Do you have any particular opinion on this question from the perspective of dotcom implementation?

>And I think it should be understood that not all of a pattern's trips will be exactly the same as far as flag stops, etc.

While I personally agree, we should probably be more explicit about this in [our own GTFS documentation](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#route_patternstxt).

⚖️ _**Next steps:**_ Given that we're nearing a 🤝  consensus on an approach and spec for canonical routes, I'm setting a deadline on this for the end of next Tuesday, 21 December. Between then and the new year I'll work on creating Asana tasks to get the implementation into the TRC backlog.

digitalcora:
Agree we should better document the "contract" of route patterns w.r.t. the trips they group together. My understanding has been that all trips in a route pattern are guaranteed to have the same 1) ordered list of stops, and 2) shape, but can vary in all other aspects — and that the purpose of the `representative_trip` relationship ~~in the API~~ is to pick one of them arbitrarily (for convenience of traversing through a has-one instead of a has-many?). We do document the relationship this way in the API:

> A trip that can be considered a canonical trip for the route pattern. This trip can be used to deduce a pattern’s canonical set of stops and shape.

**EDIT:** I realized `representative_trip_id` exists in our GTFS schema too, and we say something similar about it there:

> A trip ID that can be considered a canonical trip for the route pattern. Referring to this value in `trips.txt` and `stoptimes.txt`, a pattern's canonical set of stops and shape ID and be deduced.

losvedir:
I agree we should better document what route patterns mean. It wasn't clear to me that that was the case. One of my first [GTFS creator tickets back in the day](https://app.asana.com/0/search/1201515947680813/1154670363538321) was about creating CR route pattern IDs based on hashes of the data, and F and L stops were specified to be considered part of the hash.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,125 @@
...
+- RFC PR: [mbta/technology-docs#0006](https://github.com/mbta/technology-docs/pull/0006)
+- Asana task: [🚝 Publish canonical trips/patterns for rail routes](https://app.asana.com/0/881264583703207/1200325279789524/f)
+- Status: Proposed
+
+# Summary
+
+This will enable us to publish in our static GTFS feed "canonical" data for our rail routes. We will
+do so via a "canonical" service that's never active, but which has representative trips with stop
+times and shapes. If a canonical trip is similar enough to a real-world trip that are otherwise
+in the GTFS feed, then it will be made the representative trip of an existing route pattern. If no
```
digitalcora: From discussion in the main thread re: the nature of representative trips, it's unclear to me if there's still a case for this. @losvedir could you clarify whether you still think it's relevant to add the "fake trip" as the representative trip for existing route patterns that are made canonical? (in other words, are we talking about an as-is Concept 4️⃣, or your proposed modification?)

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,125 @@
...
+  routes. `start_date` and `end_date` should reflect the feed's valid start/end dates. All other
+  fields should have a value of `0`.
+- **calendar_attributes.txt** includes the new `service_id`. `service_schedule_type` should be
+  `"Other"`, and `rating_start_date` and `rating_end_date` should reflect the feed's valid start/end
+  dates. `service_schedule_typicality` should be `6`, and the
+  [docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md) should be updated
+  with this new kind of value.
+- **route_patterns.txt** includes new route patterns for routes/directions we want to hardcode when
+  the canonical trips (added in the trips file below) do not resemble any scheduled service. New
+  patterns in this file shall have a new `route_pattern_typicality` value of should be `6`, and
```
digitalcora: I think this should be `5` (the highest current value is `4`).

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,125 @@
...
+  `route_pattern_sort_order` should be a greater than the value for all non-canonical route patterns
+  for that given route/direction/branch. This will be done by adding two new digits to
+  `route_pattern_sort_order` (one for branch and the other for canonicalness, both placed right
+  after the existing digits corresponding to the `route_sort_order`). For routes with branches,
+  you'll need a pattern for each route/direction/branch (eg, for Providence/Stoughton Line, you
+  should have _two_ new route patterns in the outbound direction). Finally, a new
+  `canonical_route_pattern` field is added to all route patterns, indicating whether or not it can
+  be considered canonical (or if the route does not have canonical patterns defined, as will be the
```
digitalcora: We should specify the actual values here, to make it clear this is tri-state (and should probably also specify that the "no patterns defined" state is the default value if the column is blank).

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,125 @@
...
+# Prior art
+
+Until now we've been doing ad-hoc adjustments to determine the route pattern of routes
+[here](https://github.com/mbta/api/blob/master/apps/state/config/config.exs#L145).
+
+# Unresolved questions
+
+Ensuring that we would not be in a situation where the sort order, perhaps ordered by
+frequency/proportion of trips, becomes a different order than desired for the main route pattern
+against the branch route pattern.
```
digitalcora: Is this sort-of resolved now that we've specified the changes to `route_pattern_sort_order`? It's missing an explicit statement that if existing patterns are deemed canonical their relative sort order should be updated to match the intended one, but this seems feasible in principle.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,125 @@
...
+
+How route patterns containing canonical representative trips are tagged or otherwise identified:
+Using a new `canonical_route_pattern` field in route_patterns.txt, or is the new
+`service_schedule_typicality` value sufficient?
```
digitalcora: The current version of this RFC seems to resolve this question in favor of the former.

rfcs/0006-gtfs-canonical-routes.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Future possibilities
+
+We could expose this more clearly in the API. In the approach outlined here, the "canonical" route
+pattern is simply that with a typicality of `1`. In the alternative approach, the typicality would
+be `5`. That's not currently filterable in the API. In addition, the `service` can't be filtered by
+its typicality, most relevantly when included with `trips`.
```
digitalcora: This seems to be partially referring to an earlier version of the proposal.

rfcs/0006-gtfs-canonical-routes.md
jfabi: Great catch, thanks!

rfcs/0006-gtfs-canonical-routes.md
jfabi: Good idea. 👍

rfcs/0006-gtfs-canonical-routes.md
jfabi: Like with the next comment below, I wanted to check whether anyone had any qualms with the `route_pattern_sort_order` adjustment I proposed.

>It's missing an explicit statement that if existing patterns are deemed canonical their relative sort order should be updated to match the intended one, but this seems feasible in principle.

_Should_ such existing patterns have their sort order updated in some way? The only goal I had accounted for in the latest text was ensuring that canonical patterns across different branches would be sorted correctly, regardless of whether "a new fake pattern" or "an existing pattern" is utilized.

rfcs/0006-gtfs-canonical-routes.md
jfabi: It does, but I added the question myself to give folks an opportunity to question/challenge the resolution if they thought it was unnecessary.

Re your https://github.com/mbta/technology-docs/pull/6#discussion_r773282552, however, if we _don't_ make the "fake trips" representative of existing route patterns (when made canonical), then we would absolutely need a `canonical_route_pattern` field to show that canonical-ness.

rfcs/0006-gtfs-canonical-routes.md
jfabi: Yeah, I not sure if I updated this section at all. I just tried to clean it up now.

rfcs/0006-gtfs-canonical-routes.md
digitalcora: Doesn't the current approach require the field in either case? Even if the "fake trips" are always committed to the feed, you can't quite deduce the same information by looking up the service typicality of the trip (trip service typicality is 6 => pattern is canonical; trip service typicality is not 6 => ???, could be either that it is definitely-not-canonical or that no canonical patterns are defined for the route).

rfcs/0006-gtfs-canonical-routes.md
digitalcora: > The only goal I had accounted for in the latest text was ensuring that canonical patterns across different branches would be sorted correctly, regardless of whether "a new fake pattern" or "an existing pattern" is utilized.

This suggests to me the answer to your question is "yes", because there's no guarantee (is there?) that existing patterns which are marked canonical will have the same relative sort orders as the artificial canonical ones we generate. But it's very possible I'm misunderstanding you, and I think there's some vagueness in the current RFC wording:

> This will be done by adding two new digits to `route_pattern_sort_order` (one for branch and the other for canonicalness, both placed right after the existing digits corresponding to the `route_sort_order`)

Could you give a concrete example of this approach? How exactly is the digit for "branch" determined, both for our "constructed" canonical patterns and the real ones? Is the digit for "canonicity" set for constructed patterns and not set for real ones that are marked canonical, or is it set for both?

rfcs/0006-gtfs-canonical-routes.md
jfabi: You could probably infer between the two if you iterate over _all_ of the route patterns on a given route, but you're right in that it's a more tedious process. I'm more than happy to keep the `canonical_route_pattern` field, in any event.

rfcs/0006-gtfs-canonical-routes.md
jfabi: Screen Shot 2021-12-21 at 13 42 52

Above is an example I had put together when thinking through the proposed solution. (The illegal non-digits in the `route_pattern_sort_order`s I put in for temporary clarity.)

The second block (right after the "blah", which represents the corresponding `route_sort_order`) would be set in GTFS-creator based on the branch ordering, which itself would be hardcoded. For instance, for the Red Line, we could say that any pattern that touches "Savin Hill" or "Ashmont" is on the `3` branch, and any pattern touching "North Quincy" or "Braintree" is on the `4` branch.

> Is the digit for "canonicity" set for constructed patterns and not set for real ones that are marked canonical[?]

Yes on this part of the question, or at least they would be set differently, so that any exposed "constructed patterns" get sorted to the end of a branch's patterns.

⚠️ The one case where this may look a little funny is if a client is sorting route patterns exclusively on `route_pattern_sort_order`, without regard for `route_pattern_typicality` or `canonical_route_pattern`. In that case, canonical "constructed" patterns would be mixed in among in-service patterns for multi-branch routes. This is a small subset of routes, though I'd suggest mitigating this by recommending, in the [GTFS-documentation](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#route_patternstxt), to use `route_pattern_sort_order` in conjunction with the other fields, much as dotcom's existing practice.

paulswartz: Is there a reason that we should require consumers to consider other fields besides `route_pattern_sort_order` for sorting? People don't read documentation, so it would be more convenient for them (and easier for us to update in the future) if they only need to consider one field.

digitalcora: I'd also argue for `sort_order` to be all-inclusive. It's easier for developers to understand that way, and if we already know how we want clients to sort patterns, we should do it for them — that way the algorithm stays with us, instead of being encoded within (many different) clients where it's much harder to change.

jfabi: I might push back somewhat on the idea that we know how _all_ clients want to display patterns. Some clients/applications may want to show unscheduled canonical patterns, and others might not. This could (and does) occur even with our current setup, if a client wanted to obscure highly-atypical patterns into an "Uncommon destinations". But I think @paulswartz's comment stands in that, given that we don't know exactly how a given client wants to show a route's patterns, we should provide these fields as-is without recommendation as to _how_ to combine them...just as we do now.

jfabi: @digitalcora I think we can add something to GTFS-creator's truncation phase such that it checks, for these select branched routes, for these these conditions, given that a **Branch A** should be "shown" first before a **Branch B**, automatically adjusting sort orders as needed to meet them:

- Among all `canonical_route_pattern=1` patterns on the route/direction, all those on **Branch A** must have `route_pattern_sort_order`s less in value than all those on **Branch B**.
- Among all `route_pattern_typicality<5` patterns on the route/direction, at least one pattern on **Branch A** must have a `route_pattern_sort_order` less in value than the **Branch B** pattern with the smallest `route_pattern_sort_order`.
- No patterns should have the same `route_pattern_sort_order` between two different `direction_id`s on the same `route_id`.

Do these seem reasonable? If that's the case, I'll remove explicit mention of adding new digits to `route_pattern_sort_order` (if that's truly needed, especially to handle the third constraint, it can be added as part of the implementation).

rfcs/0006-gtfs-canonical-routes.md
losvedir: > @losvedir could you clarify whether you still think it's relevant to add the "fake trip" as the representative trip for existing route patterns that are made canonical?

I still do like the idea of more cleanly separating canonical data from real data. I can't shake the idea that what if someone is using a real trip in a canonical sense, but it doesn't run on the weekend, or it picks up an alert (I think CR trips can have alerts, or they are planning to do that anyway), it may come out funny.

Still, it's not necessary given the definition of route patterns as strictly just stop_ids and orders, and it's only a vague sense. I'm fine with either way that @jfabi decides.

rfcs/0006-gtfs-canonical-routes.md
losvedir: Sorry for the basic question, but can someone give me an example where `route_pattern_sort_order` is used? I'm having trouble reifying this discussion. I understand the sort order of canonical patterns for determining the "main branch", but can't think of other places I've seen all the patterns in some sorted order.

paulswartz: The website displays patterns in an order, but I don't know whether they use `route_pattern_sort_order` or not: https://www.mbta.com/schedules/9/line?schedule_direction%5Bdirection_id%5D=0&schedule_direction%5Bvariant%5D=9-_-0 (click the "City Point" dropdown to see them)

rfcs/0006-gtfs-canonical-routes.md
digitalcora: > I can't shake the idea that what if someone is using a real trip in a canonical sense, but it doesn't run on the weekend, or it picks up an alert (I think CR trips can have alerts, or they are planning to do that anyway), it may come out funny.

I think we have this problem both ways. In the world where we add the fake trip as the representative one, you'd be in trouble if e.g. you treated the times of the stops as "representative", since they're all just one minute apart (where a real trip would have realistic times). Either way, if you look at properties other than the shape and sequence of stop IDs, there is going to be something "un-representative" about the trip. I think this is an inevitable consequence of trying to fit the concept of a "canonical pattern of service" into a data model designed only to convey actual, concrete service — a solid case for inventing a separate data model a la Concept 5️⃣, though I think it's very reasonable not to go in that direction for now.

rfcs/0006-gtfs-canonical-routes.md
digitalcora: They're sorted [by typicality and then by sort order](https://github.com/mbta/dotcom/blob/74a4703/apps/route_patterns/lib/repo.ex#L44).

losvedir:
@jfabi has cleaned this up and addressed comments and started making tickets. It would be disingenuous to enter "final comment period" because we considered the time around the holidays to be that time (announced in Slack, if not here; will be better about that in the future). So I'm going to merge this PR as-is, though I will apply the "FCP" and "disposition accept" tags, for filtering and such in the future.

PR #7, opened 2021-09-14
RFC: Internal MBTA application publish vehicle IDs with a consistent prefix so as to avoid overlaps

paulswartz:


rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,185 @@
...
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to senior developers and to junior
+  developers. -->
+
+Our vehicle information is published in our GTFS-RT VehiclePositions file. There are three fields which currently contain some vehicle information:
+
+| Field | Description | Values |
+| --- | --- | --- |
+| `id` | Internal system of identification for the vehicle. Should be unique to the vehicle | Heavy/light rail: the line the car is on plus a hex code that is unique for the consist<br>Commuter Rail: the number of the control coach<br>Bus: the letter "y" plus the number of the bus |
+| `label` | a user visible label - for example the name of a train | Heavy/light rail: the number of the lead car<br>Commuter Rail: the number of the control coach<br>Bus: the number of the bus |
```
lemald: **nitpick:** For light rail, the label actually includes all cars in the consist separated by a `-` character.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,185 @@
...
+  work on this project. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to senior developers and to junior
+  developers. -->
+
+Our vehicle information is published in our GTFS-RT VehiclePositions file. There are three fields which currently contain some vehicle information:
+
+| Field | Description | Values |
+| --- | --- | --- |
+| `id` | Internal system of identification for the vehicle. Should be unique to the vehicle | Heavy/light rail: the line the car is on plus a hex code that is unique for the consist<br>Commuter Rail: the number of the control coach<br>Bus: the letter "y" plus the number of the bus |
```
lemald: **nitpick:** I think the OCS-supplied code for light rail vehicles is a decimal, rather than hexadecimal, value.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+Applications which generate vehicle IDs will need to be updated to generate
+the new IDs, but do not need to coordinate amongst themselves: no mode/line
+is generated by multiple applications.
+
+While no developers should be dependent on the existing ID formats, we should
+provide some notice (~1 month) via the Developer Google Group and our normal
+check-ins with Keolis. It is also possible that Splunk Dashboards might need
+to be updated, in particular for bus which has long used the `y` prefix.
```
ldnash: We may want to confirm that Skate is using the label field for things such as vehicle search rather than the id field. 

Let's also give Swiftly a heads-up -- I know that they use id and not label for display purposes. I don't think this would break anything, but good to let them know along with the general developer community.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,185 @@
...
+
+| Mode / Line | Prefix | Old => New |
+| --- | --- | --- |
+| Red Line | `mbta-red` | `R-546C3C33` => `mbta-red-546C3C33` |
+| Orange Line | `mbta-orange` | `O-546C3BA5` => `mbta-orange-546C3BA5` |
+| Green Line / Mattapan High-Speed Rail | `mbta-green` | `G-10100` => `mbta-green-10100` |
+| Blue Line | `mbta-blue` | `B-546C3C08` => `mbta-blue-546C3C08` |
+| Commuter Rail | `keolis-rail` | `1809` => `keolis-rail-1809`
+| MBTA Bus | `mbta-bus` | `y1454` => `mbta-bus-1454` |
+| Yankee Shuttle Bus | `yankee-bus` | `ynk1234` => `yankee-bus-1234` |
```
lemald, line 68: **suggestion (non-blocking):** not sure if it makes sense to add in this section or elsewhere, but it might be worth considering how this would work for ferry should we ever get real-time ferry location data.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,185 @@
...
+Note that having something written down in the future-possibilities section is not a reason to
+accept the current or a future RFC; such notes should be in the section on motivation or rationale
+in this or subsequent RFCs. The section merely provides additional information. -->
+
+We could consider doing something similar for trip IDs in the future. At the
+moment, MBTA and Keolis are both using HASTUS for scheduling, which uses
+integer trip IDs. They do not currently overlap, but we don't know whether
+that's a possiblity in the future. Changing the trip IDs would be a much
+larger lift however, as not all applications (TransitMaster) support
+non-integer trip IDs.
```
ldnash, line 186: Aren't TransitMaster trip IDs distinct from HASTUS/GTFS trip IDs anyway?

dks-mbta:
My thoughts: I prefer to keep things concise, and mbta-bus-1234 is a lot longer than y1234. Places with limited visual real estate like vehicle labels on the Swiftly live map would be a lot more cluttered with longer vehicle IDs.
SwiftlyLiveMap_route57

rfcs/0007-consistent-vehicle-ids.md
lemald: Good idea to check that feature in Skate, I just pulled up the code real quick and it appears that it does use label rather than ID.

digitalcora:
As a general comment, having read this, I'm unclear on what the motivation is. It says this is to avoid potential overlaps in vehicle IDs, but the current naming scheme is already non-overlapping and unambiguous (?), and there are no apparent obstacles to extending it without changing old IDs.

rfcs/0007-consistent-vehicle-ids.md
paulswartz: Yes, but we're in the process of asking TransitMaster to import our GTFS file and one of the sticking points is that we have non-integer trip IDs that they can't import.

jfabi:
> My thoughts: I prefer to keep things concise, and mbta-bus-1234 is a lot longer than y1234. Places with limited visual real estate like vehicle labels on the Swiftly live map would be a lot more cluttered with longer vehicle IDs.
> SwiftlyLiveMap_route57

We could probably ask Swiftly if they could rely instead on `label`, which would make it show as simply "1234"—since that's the purpose of the `label` field.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,186 @@
...
+| Commuter Rail | `keolis-rail` | `1809` => `keolis-rail-1809`
+| MBTA Bus | `mbta-bus` | `y1454` => `mbta-bus-1454` |
+| Yankee Shuttle Bus | `yankee-bus` | `ynk1234` => `yankee-bus-1234` |
+
+
+Applications which generate vehicle IDs will need to be updated to generate
+the new IDs, but do not need to coordinate amongst themselves: no mode/line
+is generated by multiple applications.
+
+While no developers should be dependent on the existing ID formats, we should
```
losvedir, line 75: We should probably check with OPMI / Korbato, since they have access to the OCS feed, and could potentially be using those original IDs for something. (Though since currently the realtime feed which they consume uses those numbers but prepends a line letter, unless they're doing the same I doubt they are using them).

One thing is that while we are not dependent on the existing ID format, we *are* dependent on an ID remaining constant over a trip - in particular for prediction accuracy, which uses vehicle ID to determine when a prediction is fulfilled.

So I think that means we should either accept a quick hit to prediction accuracy (not the end of the world for us, but maybe check with OPMI?), or plan to make these changes either all in one go overnight, or vehicle-by-vehicle during the day.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,186 @@
...
+only have one type of bus information. However, if we add additional sources
+of bus data (such as contracted shuttle buses) we need another prefix in
+order to distinguish the new vehicles.
+
+To avoid this, this RFC introduces consistent prefixes to give each group of vehicles a unique namespace. The pattern is
+`<vendor>-<mode/line>-<unique identifier>`.
+
+| Mode / Line | Prefix | Old => New |
+| --- | --- | --- |
+| Red Line | `mbta-red` | `R-546C3C33` => `mbta-red-546C3C33` |
```
losvedir, line 62: I'm a *little* leery with the idea of putting extraneous information in the IDs, since that means clients might end up relying on that, and changing the ID format in the future would become a breaking change (for those clients, if not by our definition).

For example, the following IDs changed recently which caused problems for various clients:
* `place-dudly` to `place-nubn`
* CR stop IDs to mile markers, e.g. `Lynn` to `place-ER-0115`
* CR trip IDs from e.g. `CR-Saturday-Fall-19-1501` to `CR-483656-506`

But I'm not sure I have a better idea. This is a pretty straightforward way for all the apps to avoid collisions, and I wouldn't want to hash it or anything like that. Maybe a line in the documentation "don't rely on the format, it may change in the future and we don't consider that a breaking change" is all that's needed.

digitalcora: [We already list](https://www.mbta.com/developers/v3-api/versioning) changes to the format or value of IDs as something we don't consider a breaking change (though this is specifically a V3 API doc). I don't think this is a totally accurate thing to say, as in the past we've definitely interpreted changes to e.g. stop IDs as a breaking change, and added backwards compatibility for them. IMO we should list each type of resource somewhere and, for each one, indicate whether changes to it are considered backwards-compatible.

lemald: I think it's also worth teasing apart two separate, but related issues, when thinking about this:
1. Clients assuming that the ID for something like a stop will remain the ID for that stop in the future.
2. Clients assuming that there is meaningful information that they can pull out of the ID rather than simply treating it as an atomic value.

I think 1 is much less of an issue for vehicles than it likely is for, say, stops and trips - clients may care about a certain stop over a long period of time, or a certain trip that happens every day, etc., but not many API / GTFS-rt consumers care about monitoring vehicles in the long term. With respect to 2, I think it's a bit of a mixed bag - I could imagine there maybe being a handful of clients that assume they can (say) remove the "y" from the beginning of a bus vehicle ID to get the number, but on rapid transit there's already nothing very meaningful about the ID aside from the R/G/B/O prefix.

losvedir: Ah, great, thanks for teasing those apart. It helped me clarify my thoughts. Consider my post here then, first and foremost, the obligatory "but should the ID be anything other than a semantically-meaningless opaque blob"?

One of the reasons meaningful IDs is bad is because the meaning could change (`dudly`) and then the ID needs to change, but as you point out, that's not really a problem here.

But the other reason meaningful names is bad is because it can encourage people to extract meaning from it, when there are better places. My `CR-Saturday-Fall-19-1501` example is relevant here, because when we changed that format we broke an internal Keolis app that relied on the ID having the rating in it.

As you point out, right now people probably aren't using the ID semantically (heh, the Swiftly livemap notwithstanding!), but when we put more info in it, we might encourage them to? I'm just picturing someone displaying vehicles, seeing `mbta-red-ABCDEF` and deciding to group vehicles by `red`, `blue`, etc.

But as Cora points out, we already say that's not something we promise to maintain, so maybe we just 💧 ✋ of the situation and if someone uses it incorrectly it's on them if we decided to change it.

rfcs/0007-consistent-vehicle-ids.md
```diff
@@ -0,0 +1,186 @@
...
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to senior developers and to junior
+  developers. -->
+
+Our vehicle information is published in our GTFS-RT VehiclePositions file. There are three fields which currently contain some vehicle information:
+
+| Field | Description | Values |
+| --- | --- | --- |
+| `id` | Internal system of identification for the vehicle. Should be unique to the vehicle | Heavy rail: the line the car is on plus a hex code that is unique for the consist<br>Light rail: "G-" plus a decimal code that is unique for the consist<br>Commuter Rail: the number of the control coach<br>Bus: the letter "y" plus the number of the bus |
+| `label` | a user visible label - for example the name of a train | Heavy rail: the number of the lead car<br>Light rail: the numbers of each car in the consist, separated by `-`<br>Commuter Rail: the number of the control coach<br>Bus: the number of the bus |
```
skyqrose, line 48: nitpick: For Commuter Rail, when a locomotive is at the front of the train, it'd be the number of the "locomotive" rather than "control coach", right?

rfcs/0007-consistent-vehicle-ids.md
skyqrose: Adding on to that: people already can parse out the route out of a heavy rail id, but it feels wrong, like you're not supposed to do it and you're cheating a bit, and you're depending on clearly undocumented behavior. Having human-readable ids makes it feel more trustworthy and documented, even if it isn't intended to be.

One way to have non-overlapping ids without encouraging use would be to have semantic but not-very-readable ids. The one letter prefixes we have right now are great at that, even if they're not a consistent format between the modes.

I'd propose that we change commuter rail ids to be something like `rr1809` (for "regional rail" or "rail road") so they're namespaced, and avoid having to change other already-namespaced modes and without giving the impression that it's meant to be parsed.

skyqrose:
How are ferry ids currently structured?

rfcs/0007-consistent-vehicle-ids.md
lemald: IIRC the label is associated with the GPS tracking unit, and there's only one of those per train, which is always in the control coach, regardless of which direction it's heading.

lemald:
> How are ferry ids currently structured?

We don't currently have real-time data for ferry so I don't believe there is any place in our systems that has a ferry vehicle ID.

rfcs/0007-consistent-vehicle-ids.md
lemald: > I'd propose that we change commuter rail ids to be something like rr1809 (for "regional rail" or "rail road") so they're namespaced, and avoid having to change other already-namespaced modes and without giving the impression that it's meant to be parsed.

This still isn't as namespaced as we want - it indicates that it's commuter rail, but not that it's Keolis (what happens if we ever have commuter rail service operated by more than one entity?). Trying to abbreviate all of that information down seems like it will just end up more confusing in the end. My overall opinion is that we should focus on better documenting how consumers can obtain all of this data correctly rather than obfuscating data sources that we don't want them to use

digitalcora: Hot take: all vehicle IDs are UUIDv5, the operator name is the namespace and the original vehicle ID is the name. Try getting any meaningful information out of that 😄 (yes, I know it's often useful for _us_ specifically to be able to look at a plain vehicle ID in log output or something and see that it's a Red Line car)

jfabi:
> > How are ferry ids currently structured?
> 
> We don't currently have real-time data for ferry so I don't believe there is any place in our systems that has a ferry vehicle ID.

The two ferries we have GPS tracking for—but currently no programmatic way of getting the data out—have internal IDs of `Champion - 0010f34f8ca5` and `Glory - 0010f350899f`. But the boats also have three-digit identifiers that we might be able to use instead.

Either way, I don’t think ferry should play a big factor into which format is used. 🙂

paulswartz:
Thank you everyone for your feedback! 

Given the lack of a concrete use-case for this standardization, I'm going to close out this RFC. If we have a use-case in the future, we can bring it back.

PR #8, opened 2021-09-25
rfc: Commuter rail crowding

losvedir:
An RFC about how to process Keolis's upcoming commuter rail crowding data for dotcom (and potentially others) to use.

### DIFF
--- a/rfcs/0000-api-cr-crowding.md
+++ b/rfcs/0000-api-cr-crowding.md
@@ -1,0 +1,337 @@
+- Feature Name: api_cr_crowding
+- Start Date: 2021-09-23
+- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
+- Asana task: [asana link](https://app.asana.com/)
+- Status: Proposed
+
+# Summary
+
+[summary]: #summary
+
+We want to provide predicted commuter rail crowding data for upcoming trips. The data will be
+consumed from a feed provided by Keolis. We will follow the lead of this [static GTFS-Occupanices
+proposal](https://github.com/google/transit/pull/240), but for now the data will only appear in
+the API and not in our GTFS feed. The API will gain a new "occupancies" resource. The data will be
+documented in the API's swagger docs as "Experimental" with a warning that it could change in the
+future. The implementation will be as minimal and simple as possible, but nevertheless built in a
+solid and (hopefully) forward-compatible way. The immediate concrete use case for this feature is
+showing crowding information on dotcom's CR timetables pages.
+
+# Motivation
+
+[motivation]: #motivation
+
+This data feed will allow clients to easily see expected crowding for CR trips. In particular,
+dotcom is planning to add this information to the CR timetable pages.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+<!-- Explain the proposal as if it was already implemented and you were teaching it to a new developer
+that just joined the team. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how programmers should _think_ about the feature, and how it should impact the way they
+  work on this project. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to senior developers and to junior
+  developers.
+-->
+
+This introduces into the API the proposed new GTFS static resource called
+[occupancies.txt](https://github.com/google/transit/pull/240/files), which contains information
+about "usual or expected in-vehicle occupancy levels". The underlying data can apply to either a
+whole trip or individual stop times along the trip. A given occupancy specifies its
+`occupancy_status` which is an enum ranging from `0` for "Empty" to `6` for "Full". Similar to how
+we represent vehicles in the API, the API response will be the enum word rather than the number.
+Our feed will have an _extra_ field, currently not part of the spec, but which is being advocated
+for, `occupancy_percentage`, which is a value from `0.0` to `1.0`. Each occupancy has some
+validity constraints, namely the start and end dates and days of the week for which they're valid.
+A trip could therefore have several related occupancies.
+
+Our Keolis source data is at the trip level, and that is the only producer for which we will have
+static occupancy data in the near term. Initially, we will put this data directly in the API and
+not in our GTFS feed since the data updates as frequently as once/hour, and we don't want to
+update our GTFS static feed that often. However, we should ensure the API interface works _as if_
+it were coming from the GTFS feed, in case we do so in the future. While we're not given this
+data, we can arbitrarily say that the occupancies are valid for two weeks and apply to all days of
+the week. A weekend train trip generally doesn't run on weekdays, but if it were to (e.g. Labor
+Day running Sunday service), we'd want the predicted occupancy to be considered valid.
+
+Our API largely matches the GTFS feed and is simply a convenient way of accessing that data. While
+there are a few GTFS resources that we use in the API that are not exposed directly (e.g.
+transfers, calendar) and one major one that's named differently (schedule vs stoptime), the
+default expectation is that GTFS resources are available in the API.
+
+To that end, the API has a new `/occupancies` endpoint for an `Occupancy` resource. While the
+occupancy must have an ID – which we'll construct as `occupancy-[trip-id]` – there's no
+`/occupancies/[id]` endpoint, akin to `/predictions` or `/schedules`.
+
+For now, due to the experimental and malleable nature of the proposal, we want to start small and
+limit the coupling of occupancies with the rest of our code. So at a minimum we will provide just
+a single `/occupancies` endpoint that returns all current occupancies, with their trip
+relationship, like such:
+
+```
+GET /occupancies
+
+{
+  data: [
+    {
+      "attributes": {
+        "occupancy_status": MANY_SEATS_AVAILABLE,
+        "occupancy_percentage": 0.11,
+        "start_date": [today],
+        "end_date": [2 weeks from today],
+        "monday": 1,
+        "tuesday": 1,
+        "wednesday": 1,
+        "thursday": 1,
+        "friday": 1,
+        "saturday": 1,
+        "sunday": 1
+      }
+      "id": occupancy-[trip-id]-all,
+      "relationships": {
+        "trip": {
+          "data": {
+            "id": trip_id,
+            "type": "trip"
+          }
+        },
+      },
+      "type": "occupancy"
+    },
+    // ...
+  ]
+}
+```
+
+From there we will add the minimum extra functionality required for our one known consumer,
+dotcom. This _might_ be enough as-is. I could foresee an implementation where they periodically
+query this endpoint, store it indexed by trip ID and that's enough for the timetable page.
+
+However, I could imagine a couple extensions they might need:
+
+- `?filter[route]=CR-route1`, `?filter[trip]=trip1,trip2,trip3`, or
+  `?filter[trip.name]=name1,name2,name3`. I'm not sure how the timetables work, but maybe querying
+  only relevant occupancies via route, trip, or trip short name is necessary.
+- `?include=trip`. Maybe filtering isn't necessary, but being able to include the trip, in order
+  to have easy access to the trip short name, is necessary.
+
+But for the time being, we won't allow other resources to include occupancies. For example, we
+don't want clients to do something like `/trips?filter[route]=CR-Worcester&include=occupancies`.
+In the event that something substantial changes about occupancies (or if it's removed!), the
+changes are limited to just this one resource, which is a risk of using the experimental resource.
+On the other hand, if this proposal matures and we build forth, clients who have built
+interactions with this endpoint will continue to have this as a supported method of access.
+
+# Reference-level explanation
+
+<!--[reference-level-explanation]: #reference-level-explanation
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how
+the detailed proposal makes those examples work. -->
+
+## Keolis data
+
+The Keolis data comes from a firebase feed. It looks like:
+
+```
+{
+  "data": [
+    {
+      "cTrainNo": "003     ",
+      "MedianDensity": 0.58635,
+      "MedianDensityFlag": 1
+    },
+    //...
+  ],
+  "last_updated": "2021-09-24 15:30:00"
+}
+```
+
+Each record in `"data"` is for one trip, with `"cTrainNo"` being the train's name (with extraneous
+padding for arcane reasons on Keolis's end).
+
+The `"MedianDensity"` is the percentage of seats occupied, and `"MedianDensityFlag` is an enum
+with values `0`, `1`, or `2`, representing Keolis's idea of "Not Crowded", "Some Crowding", and
+"Crowded" respectively.
+
+## API data
+
+In the API, we'll add a new kind of model `Occupancy` with its own `State.Server`. The model will
+have keys (following the GTFS proposal):
+
+```
+[
+  :trip_id,
+  :stop_sequence,
+  :occupancy_status,
+  :occupancy_percentage,
+  :valid_days,
+  :start_date,
+  :end_date,
+  :start_time
+]
+```
+
+- The `:trip_id` can be looked up using a trimmed version of `cTrainNo` which is equivalent to a
+  trip's GTFS `trip_short_name`.
+- `:stop_sequence` will be `nil`
+- `:occupancy_status` can in theory be one of `:empty`, `:many_seats_available`,
+  `:few_seats_available`, `:standing_room_only`, `:crushed_standing_room_only`, `:full`, or
+  `:not_accepting_passengers`, but for the Keolis data will only use 3 of them, with the mapping
+  based on `MedianDensityFlag` of `0` to `:many_seats_available`, `1` to `:few_seats_available`,
+  and `2` to `:full` (unless dotcom already has a different mapping of GTFS terms to bowling
+  pins?).
+- `:occupancy_percentage` is just the `MedianDensity` value verbatim.
+- `:valid_days` is the list `[1, 2, 3, 4, 5, 6, 7]` representing this data applying to all days of
+  the week (akin to how `Service` works in the API).
+- `:start_date` is the current date
+- `:end_date` is the date two weeks from today
+- `:start_time` is `nil`.
+
+The index (/indices) will depend on what sort of filtering (if any) we provide, based on above.
+
+## Fetching the data
+
+The Keolis endpoint should be fetched every ten minutes. We could use with modifications the
+existing `StateMediator` infrastructure to do this. State mediators can be configured to fetch
+from a URL on a periodic basis, but they take `url` as a string configuration, while the URL for
+the firebase feed changes periodically because it has a `?token=....` parameter, and that token is
+only valid for a short while. So `StateMediator.Mediator` should be updated to optionally take a
+MFA tuple for its URL configuration as well.
+
+From futzing around in a Livebook, some code to download the data from Keolis is:
+
+```ex
+Mix.install([:jason, :hackney, goth: "1.3.0-rc.3"])
+firebase_url = "..."
+credentials_json = "..."
+credentials = Jason.decode!(credentials_json)
+
+scopes = [
+  "https://www.googleapis.com/auth/firebase.database",
+  "https://www.googleapis.com/auth/userinfo.email"
+]
+
+source = {:service_account, creds, scopes: scopes}
+{:ok, pid} = Goth.start_link(name: MyGoth, source: source)
+{:ok, token} = Goth.fetch(MyGoth)
+resp = :hackney.request(firebase_url <> ".json?access_token=" <> token.token)
+{:ok, 200, _, ref} = resp
+{:ok, data} = :hackney.body(ref)
+Jason.decode!(data)
+```
+
+The `Mediator` can invoke `new_state` on its configured module (`State.Occupancy`), which will
+then do the mapping and merging of the data as specified above.
+
+## API controller
+
+Fill out the swagger docs, but make sure it says **EXPERIMENTAL** and "can change without prior
+warning".
+
+The `includes` and `filters` sections depend on what is decided above, about what is necessary for
+dotcom.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+A drawback of building on GTFS-Occupancies at this point, is that it's still in flux and can
+change in the future. But we think that it's better to build on what's there, which will hopefully
+be close to the final product (it's gone through lots of revisions at this point and seems to be
+settling down) rather than do our own thing and then end up with a totally different way of
+handling this data in the future.
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Actual GTFS
+
+The proposal we are modeling after is a static GTFS file `occupancies.txt`. An approach that we
+could have taken is for `gtfs_creator` to use the Keolis endpoint in its build process, and then
+the API to download the data via the GTFS feed.
+
+The main reason to avoid this is the data changes too frequently, on the order of once/hour, while
+we don't usually publish a feed more than once or twice a day. Plus the frequent changes would
+make it harder for clients (who aren't using `occupancies.txt` yet) to know if the data they care
+about has changed.
+
+## Small, temporary endpoint
+
+We could have done a hacky temporary `/temp_commuter_rail_crowding` endpoint. This would be low
+effort, and is appropriately quarantined from the rest of the code.
+
+However, this would set an unhappy precedent about "temporary" endpoints just for dotcom that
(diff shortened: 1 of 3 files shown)


rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+        "end_date": [2 weeks from today],
+        "monday": 1,
+        "tuesday": 1,
+        "wednesday": 1,
+        "thursday": 1,
+        "friday": 1,
+        "saturday": 1,
+        "sunday": 1
+      }
+      "id": occupancy-[trip-id]-all,
```
digitalcora: I don't think this `-all` is explained elsewhere? The details above just say that the IDs will be in the form `occupancy-[trip_id]`.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+
+```
+GET /occupancies
+
+{
+  data: [
+    {
+      "attributes": {
+        "occupancy_status": MANY_SEATS_AVAILABLE,
+        "occupancy_percentage": 0.11,
```
digitalcora: Since the resource _is_ an "occupancy" already, I'd argue these could just be `status` and `percentage`. We don't name the start date field `occupancy_start_date`, for example.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+        "start_date": [today],
+        "end_date": [2 weeks from today],
+        "monday": 1,
+        "tuesday": 1,
+        "wednesday": 1,
+        "thursday": 1,
+        "friday": 1,
+        "saturday": 1,
+        "sunday": 1
```
digitalcora: This seems like it might not be the most convenient way to represent this data for API consumers — it's just a 1:1 translation of the GTFS columns. At the least, this is JSON, so we have real boolean `true` and `false`.

In most languages it's easier to deal with "days of week" as numbers than as names, so we could provide a list of day numbers, like:

```json
"start_date": "2021-01-01",
"end_date": "2021-01-31",
"days_of_week": [1, 2, 3, 4, 5],
```

...being careful to specify whether these are 1-indexed or 0-indexed, of course.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+## Actual GTFS
+
+The proposal we are modeling after is a static GTFS file `occupancies.txt`. An approach that we
+could have taken is for `gtfs_creator` to use the Keolis endpoint in its build process, and then
+the API to download the data via the GTFS feed.
+
+The main reason to avoid this is the data changes too frequently, on the order of once/hour, while
+we don't usually publish a feed more than once or twice a day. Plus the frequent changes would
+make it harder for clients (who aren't using `occupancies.txt` yet) to know if the data they care
+about has changed.
```
digitalcora: This almost seems like an argument against the premise of this RFC, which is building an API that _could_ be powered by GTFS-Occupancies. If we don't expect it will (ever?) be practical to update GTFS as frequently as we'd like, it seems like we should just do it however's convenient in the API, without regard for how the Occupancies proposal is structured (making this an API-only feature).

That said, I'm unclear on why we'd want to update occupancy data this frequently anyway. It doesn't seem like the Occupancies proposal anticipates this, since it's part of static GTFS which assumes a "download it once in a while, idk" update frequency. Additionally, a different section of this RFC specifies we should tell clients these values are good for two whole weeks.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+
+- The `:trip_id` can be looked up using a trimmed version of `cTrainNo` which is equivalent to a
+  trip's GTFS `trip_short_name`.
+- `:stop_sequence` will be `nil`
+- `:occupancy_status` can in theory be one of `:empty`, `:many_seats_available`,
+  `:few_seats_available`, `:standing_room_only`, `:crushed_standing_room_only`, `:full`, or
+  `:not_accepting_passengers`, but for the Keolis data will only use 3 of them, with the mapping
+  based on `MedianDensityFlag` of `0` to `:many_seats_available`, `1` to `:few_seats_available`,
+  and `2` to `:full` (unless dotcom already has a different mapping of GTFS terms to bowling
+  pins?).
```
digitalcora: Since we already do this for real-time crowding, maybe we could just say here that the values should be the same as they are there.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+- `:occupancy_percentage` is just the `MedianDensity` value verbatim.
+- `:valid_days` is the list `[1, 2, 3, 4, 5, 6, 7]` representing this data applying to all days of
+  the week (akin to how `Service` works in the API).
+- `:start_date` is the current date
+- `:end_date` is the date two weeks from today
+- `:start_time` is `nil`.
```
digitalcora: Any reason this or `stop_sequence` should be present initially?

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+# Summary
+
+[summary]: #summary
+
+We want to provide predicted commuter rail crowding data for upcoming trips. The data will be
+consumed from a feed provided by Keolis. We will follow the lead of this [static GTFS-Occupanices
+proposal](https://github.com/google/transit/pull/240), but for now the data will only appear in
+the API and not in our GTFS feed. The API will gain a new "occupancies" resource. The data will be
+documented in the API's swagger docs as "Experimental" with a warning that it could change in the
+future. The implementation will be as minimal and simple as possible, but nevertheless built in a
```
digitalcora: > The data will be documented in the API's swagger docs as "Experimental" with a warning that it could change in the future.

In practice, what is the compatibility story going to be for this endpoint? We know there will be at least one important consumer who depends on this data in production — any further changes to it _would_ have to be made in a backward-compatible way, on pain of breaking the web site. I don't know if there'd be any harm in applying our usual versioning strategy to it, but this does mean it would be no more "experimental" than any other endpoint; if you kept using the same version of the API, your app wouldn't break.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+
+- What filters and includes does dotcom absolutely need on the `/occupancies` endpoint, to do
+  their work? Can they get by with a simple, unfiltered `/occupancies` with the data and the trip
+  ID in it?
+- Would dotcom strongly prefer retrieving `occupancies` via an `include` on a trip instead?
+- Does dotcom already have an established mapping of GTFS crowding terms to bowling pins? We
+  should be in line with that in our translation of Keolis's "not crowded", "some crowding",
+  "crowded" flags to GTFS terms.
+
+# Future possibilities
```
digitalcora: I'd like to see us address here how we might handle stop-time-level crowding data, and how this RFC supports that or doesn't. The GTFS-Occupancies proposal that's the foundation of this RFC is based around crowding data for specific stop sequences (it doesn't even support trip-level crowding explicitly), and we already have a known potential use case for this in the form of our subway crowding data.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+
+The GTFS-compatible way of attaching a list of occupancies to a trip in the API would be to still
+have the `Occupancy` resource, and allow `trips` to `include=occupancies`. I'd be open to this
+idea, if dotcom thought it would be the most natural way to work with the data, though I'm
+hesitant about it because I'm not exactly sure of the joining logic here, given the data model
+issues mentioned before. Should we validate date range and days of the week as part of the join?
+My gut feeling is that an isolated `/occupancies` endpoint would be simpler for us to build and
+maintain while the proposal is being worked on. Having to work with occupancies via trips also
+seems like it would couple client implementations between their standard trips and this
+non-standard feature more.
```
digitalcora: Put this way, `/occupancies` definitely seems like a "transitional" implementation that's oriented more towards our own convenience than anything else. Occupancy data pretty much has to be connected to trips or stop times for it to be meaningful, and with this approach, clients are forced to put the pieces together themselves.

We say this is a trade-off for making the data "experimental" and subject to wholesale change or removal, but with my other comment about compatibility for this endpoint, that argument seems less clear — there's some level of support we'd have to provide in either case. _Does_ a standalone endpoint really preserve some level of flexibility that a relationship doesn't?

digitalcora (commented):
I had some Comments, as Requested 😄 Overall I do think this is a better direction than `/commuter_rail_crowding` and really appreciate you laying out the details here.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+}
+```
+
+From there we will add the minimum extra functionality required for our one known consumer,
+dotcom. This _might_ be enough as-is. I could foresee an implementation where they periodically
+query this endpoint, store it indexed by trip ID and that's enough for the timetable page.
+
+However, I could imagine a couple extensions they might need:
+
+- `?filter[route]=CR-route1`, `?filter[trip]=trip1,trip2,trip3`, or
```
thecristen: The route one (and maybe additionally direction) is the most relevant extension that might be useful, but we can do without it.

rfcs/0000-api-cr-crowding.md
thecristen: Dotcom's parsing:

https://github.com/mbta/dotcom/blob/337569cc6a942773f45390be2884a623b52e9ef9/apps/vehicles/lib/parser.ex#L53-L58

```elixir
@spec crowding(String.t()) :: Vehicle.crowding()
defp crowding("MANY_SEATS_AVAILABLE"), do: :not_crowded
defp crowding("FEW_SEATS_AVAILABLE"), do: :some_crowding
defp crowding("FULL"), do: :crowded
defp crowding(_), do: nil
```

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+{:ok, data} = :hackney.body(ref)
+Jason.decode!(data)
+```
+
+The `Mediator` can invoke `new_state` on its configured module (`State.Occupancy`), which will
+then do the mapping and merging of the data as specified above.
+
+## API controller
+
+Fill out the swagger docs, but make sure it says **EXPERIMENTAL** and "can change without prior
```
thecristen: This might be scope creep but I wonder if experimental flags are something that can be enabled/disabled per MBTA API account? Therefore its use can be limited to certain internal users?

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+
+And the most likely outcome of this "experiment" is that it lingers indefinitely. We don't want a
+"temporary", ad-hoc endpoint perpetually in the API, especially when its functionality overlaps
+with a GTFS standard that will (eventually) be accepted, and we'll want to adopt formally, for
+Transit app, Google Maps, etc.
+
+## Not exposed directly, but joined to other models
+
+Another approach is to not have an `/occupancies` endpoint but still have the model and data,
+either simply stuck in a `Trip`, or `include=occupancies` by a trip.
```
thecristen: This almost seems the most elegant to me!

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+reasoning given is that a given trip could have different occupancy predictions for different date
+spans or days of the week. While we could handle this in the API via giving trips a _list_ of
+occupancies, this isn't feasible in `trips.txt` in GTFS. And so in the name of forward
+compatibility with the proposal once it lands, we shouldn't do it that way now.
+
+The GTFS-compatible way of attaching a list of occupancies to a trip in the API would be to still
+have the `Occupancy` resource, and allow `trips` to `include=occupancies`. I'd be open to this
+idea, if dotcom thought it would be the most natural way to work with the data, though I'm
+hesitant about it because I'm not exactly sure of the joining logic here, given the data model
+issues mentioned before. Should we validate date range and days of the week as part of the join?
```
thecristen: Are trip IDs not unique to certain dates/days? 🤔

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+[prior-art]: #prior-art
+
+n/a
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- What filters and includes does dotcom absolutely need on the `/occupancies` endpoint, to do
+  their work? Can they get by with a simple, unfiltered `/occupancies` with the data and the trip
```
thecristen: FWIW I do think we can get by with that.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+-->
+
+This introduces into the API the proposed new GTFS static resource called
+[occupancies.txt](https://github.com/google/transit/pull/240/files), which contains information
+about "usual or expected in-vehicle occupancy levels". The underlying data can apply to either a
+whole trip or individual stop times along the trip. A given occupancy specifies its
+`occupancy_status` which is an enum ranging from `0` for "Empty" to `6` for "Full". Similar to how
+we represent vehicles in the API, the API response will be the enum word rather than the number.
+Our feed will have an _extra_ field, currently not part of the spec, but which is being advocated
+for, `occupancy_percentage`, which is a value from `0.0` to `1.0`. Each occupancy has some
```
paulswartz: `occupancy_percentage` in the VehiclePositions feed is an integer,`0` to `100` (or more)

rfcs/0000-api-cr-crowding.md
paulswartz: That's fair, but we use `occupancy_status` in other parts of the API / GTFS-RT.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+relationship, like such:
+
+```
+GET /occupancies
+
+{
+  data: [
+    {
+      "attributes": {
+        "occupancy_status": MANY_SEATS_AVAILABLE,
```
paulswartz:
```suggestion
        "occupancy_status": "MANY_SEATS_AVAILABLE",
```

rfcs/0000-api-cr-crowding.md
paulswartz: What about doing this closer to how we do schedules, and provide a `filter[date]` parameter rather than the raw data?

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+change in the future. But we think that it's better to build on what's there, which will hopefully
+be close to the final product (it's gone through lots of revisions at this point and seems to be
+settling down) rather than do our own thing and then end up with a totally different way of
+handling this data in the future.
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Actual GTFS
```
paulswartz: What about an approach where an app generates an `occupancies.txt` separately from our GTFS, and the API consumes that file directly?

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+My gut feeling is that an isolated `/occupancies` endpoint would be simpler for us to build and
+maintain while the proposal is being worked on. Having to work with occupancies via trips also
+seems like it would couple client implementations between their standard trips and this
+non-standard feature more.
+
+# Prior art
+
+[prior-art]: #prior-art
+
+n/a
```
paulswartz: The IBM parking feed is a custom parser which is part of the API.

rfcs/0000-api-cr-crowding.md
paulswartz: This ID doesn't support having the same trip with multiple occupancies (say for example if there's an updated occupancy estimate for the future but not today. Maybe include the `start_date`?

rfcs/0000-api-cr-crowding.md
digitalcora: With the `occupancy_status` on `Vehicle` resources, I think the prefix makes more sense because the "occupancy" is not the resource the field is attached to (if it were just `vehicle.status` it would be ambiguous).

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- What filters and includes does dotcom absolutely need on the `/occupancies` endpoint, to do
+  their work? Can they get by with a simple, unfiltered `/occupancies` with the data and the trip
+  ID in it?
+- Would dotcom strongly prefer retrieving `occupancies` via an `include` on a trip instead?
```
thecristen: It sounds nice but I think it's ok if we don't have that.

rfcs/0000-api-cr-crowding.md
```diff
@@ -0,0 +1,337 @@
...
+- Does dotcom already have an established mapping of GTFS crowding terms to bowling pins? We
+  should be in line with that in our translation of Keolis's "not crowded", "some crowding",
+  "crowded" flags to GTFS terms.
```
thecristen: See comment above.

thecristen (commented):
Assorted comments ✨

rfcs/0000-api-cr-crowding.md
losvedir: Oops, I had an earlier version which specified `-all` for trip level occupancies and `-[stop_sequence]` if we had them. Then I realized that we don't have those other ones, and clients aren't really going to rely on a continuity of ID, so it's not a problem to change the form later if we need to disambiguate. But then I missed the code snippet here. Thanks.

rfcs/0000-api-cr-crowding.md
losvedir: Ah, this is interesting. I initially just copied the name of the column from the GTFS proposal, and I do like that it's the same as on Vehicle.

However, this prompted me to look at how we do it elsewhere in the API, and it seems to be generally the case that the GTFS field repeats the resource name, but the API does not include it.

E.g:

## stops

**GTFS**
- stop_name
- stop_desc
- stop_lat
- stop_lon

**API**
- name
- description
- latitude
- longitude

## trips

**GTFS**
- trip_headsign
- trip_short_name

**API**
- headsign
- name

So I think in light of the precedent here, I'll omit the resource name from the field and go with `status` and `percentage`.

rfcs/0000-api-cr-crowding.md
losvedir: These are both good approaches. I think what this means is I'll punt on the question. We're not actually getting this data from Keolis, and we don't expect any client to use it at the moment, so no need to commit to either approach right now. I'll just remove this from the API for now.

rfcs/0000-api-cr-crowding.md
losvedir: I just mean right now we're not set up to publish that frequently. It's conceivable we'd have a more frequently updating feed in the future. We'd have to tweak our feed archival logic, and make sure there's a good way for clients to know what changed. Or maybe have a frequently updating feed and a more significant snapshot feed. Or a separate `occupancies.txt` file adjacent to the feed. Or maybe the data *doesn't* meaningfully change that often and we can get by with a daily or less release. (I feel like this last idea is the most likely, but for the moment Keolis is hot to trot on the up-to-date nature of the data here and wants to make sure it's available to clients like this.)

All this to say, I don't think it's a long term deal breaker, just not something I want to deal with right now. But you're right that the RFC is kind of phrased that way, so I'll update it with some of this comment.

rfcs/0000-api-cr-crowding.md
losvedir: Sure, I'll say that. But since Cristen confirmed the values and the prompt for this section is "it is reasonably clear how the feature would be implemented", I figure might as well mention the actual values here, too.

rfcs/0000-api-cr-crowding.md
losvedir: Nope!

rfcs/0000-api-cr-crowding.md
losvedir: I think that because dotcom is internal to our organization we at least have the potential to tweak things (either by asking @thecristen very nicely and offering her a tasty coffee drink, or issuing the PR ourselves). So I'd like to reserve the right to make changes to the endpoint and not have to maintain previous versions. For example, we're passing along the `percentage` field for now, at Keolis's request and per their suggestion in the GTFS proposal. But if it doesn't make it and dotcom uses a design that doesn't expose that value, I'd like to be able to just remove that field from our data and response.

Versioning is possible, but adds overhead. Plus, there's the additional complication that dotcom isn't on the latest version now anyway, so they wouldn't really be able to easily update to use a newer variant of this endpoint.

rfcs/0000-api-cr-crowding.md
losvedir: Good point. I'll add some thoughts here.

rfcs/0000-api-cr-crowding.md
digitalcora: A thought: clients should have to explicitly opt into behavior that might have unannounced breaking changes. I think the most flexible way to do this would be via an HTTP header, which should also be familiar for clients since they already must do this to provide their API key. For example: `x-enable-experimental-features: true`. A plug could pick this up and put a boolean value into `conn.assigns[:experimental?]` or something.

This should work regardless of whether the data is provided as a standalone endpoint or a relationship, so we could choose the approach that makes the most sense for clients. It would also provide a reusable framework for implementing other "experimental" features (though maybe we shouldn't make a habit of it).

rfcs/0000-api-cr-crowding.md
losvedir: > Occupancy data pretty much has to be connected to trips or stop times for it to be meaningful, and with this approach, clients are forced to put the pieces together themselves. ... Does a standalone endpoint really preserve some level of flexibility that a relationship doesn't?

I think you're right that it actually doesn't. I was thinking from the API side we would want to punt on questions like "well, should we make trip <-> occupancy 1-to-1 or 1-to-many?". And while forcing clients to go to `/occupancies` does do that, it just means the client now has to make that decision when, e.g., grouping the list by trip ID. And that's a much more subtle data dependency. Changing the data model by, e.g., adding multiple occupancies per trip via stop sequence would be just as breaking for dotcom as changing `include=occupancy` to `include=occupancies`, for example.

Combined with @thecristen thinking the `include` approach is more elegant, let's go with including via `trips` and *not* via an `/occupancies` endpoint. That would get dotcom the ability to filter by route and direction_id for free, both of which "might be useful".

rfcs/0000-api-cr-crowding.md
losvedir: I like that idea!

rfcs/0000-api-cr-crowding.md
losvedir: I thought about that. But *technically* even including the `start_date` isn't enough since we could use a scheme with the same start/end dates, and different valid days, to, e.g., distinguish weekday and weekend travel. I'm not sure how to unequivocally get a unique ID for all possible kinds of the data, short of just repeating all the data in the ID! (or hashing it).

But I don't think clients will really rely that much on the continuity of the ID, so I figured it would be fine to just use something that works now with the data we've got, and change the shape later, if need be.

rfcs/0000-api-cr-crowding.md
losvedir: Paul had suggested something like that, too. That's a neat idea but I think too big of a lift for us right now. But, cora suggested a smaller thing: requiring a special header to opt in. That's easy to implement it, and now it makes the client explicitly opt in and (hopefully) know what they're doing. So we're going to go with that for now. If we have more experimental stuff in the future (though, hopefully not), we might work out a heavier weight thing like having per-account flags.

rfcs/0000-api-cr-crowding.md
losvedir: I'm sold! This comment plus others have swayed me to switch to this approach. That means you'll get the filter by route and direction ID for free, too. Should be a smaller implementation on our end, too. So wins all around. 🎉

rfcs/0000-api-cr-crowding.md
losvedir: Trips are unique per *service*, and the service specifies what days the trip runs. So, e.g., the trip [`CR-483656-506`](https://api-v3.mbta.com/trips/CR-483656-506) is on the [`SummerWeekday`](https://api-v3.mbta.com/services/SummerWeekday) service. So we know that trip runs  from 9/21 to 12/17 Monday-Friday.

But the occupancies proposal allows further filtering by dates or days of the week. For example, maybe Mondays are expected to be extra busy or Fridays less crowded. Or you could specify different start/end dates for the *occupancy* that are *different* from the service, maybe to earmark certain holidays as especially busy or not busy.

Now the Keolis data isn't doing any of that right now, so this is more just a theoretical concern about how we should commit to joins happening and such. In fact, I've dropped "valid days" (though not start/end date) from our data model entirely at the moment.

rfcs/0000-api-cr-crowding.md
losvedir: "sounds nice" you say? Well, /gif you got it, dude. 👍

losvedir:
Thanks for the comments everyone! I think I've addressed everything. The major changes are:

* Switching from `/occupancies` to `/trips?include=occupancies`.
* Gating it by a special `x-include-experimental-features` header.

losvedir:
Since I believe I've addressed and either responded to or incorporated the feedback from all the comments, I'm going to move this to "final comment period" with a disposition to accept, which I'll do later today.

PR #9, opened 2021-10-06
rfc: Arrow Cognito user roles

losvedir:
Just a quick RFC about an approach to manage authorization roles in Arrow, via Cognito user groups.

Not a full RFC with all the sections, so we might end up closing it, but I wanted to document a rough sketch of what we could do, and figured this could be a good spot for it rather than a gist or something that could be lost.

### DIFF
--- a/rfcs/0000-arrow-user-roles.md
+++ b/rfcs/0000-arrow-user-roles.md
@@ -1,0 +1,168 @@
+- Feature Name: (fill me in with a unique ident, `my_awesome_feature`)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
+- Asana task: [asana link](https://app.asana.com/)
+- Status: Proposed
+
+# Summary
+
+[summary]: #summary
+
+This is a proposed method of adding user roles and permissions levels to Arrow, perhaps to be used
+by other CTD apps.
+
+# Motivation
+
+[motivation]: #motivation
+
+Arrow is currently only accessible to admins, but we want a read-only view for "viewers", and will
+likely have other roles in the future.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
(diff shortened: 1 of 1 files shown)


rfcs/0000-arrow-user-roles.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+```ex
+defmodule ArrowWeb.Plug.LoadUser do
+  import Plug.Conn
+
+  @cognito_groups %{
+    "arrow-viewer" => :viewer,
+    "arrow-admin" => :admin
+  }
```
paulswartz, line 120: suggestion: rather than configuring them here, I'd put them in `config/config.exs` (and then maybe load them here with `Application.compile_env!`

rfcs/0000-arrow-user-roles.md
```diff
@@ -0,0 +1,168 @@
...
+questions.
+
+```ex
+defmodule Arrow.Permissions do
+  alias Arrow.Accounts.User
+
+  @required_groups [
+    delete_bar: [:admin],
+    do_foo: [:admin, :viewer],
+  ]
```
paulswartz, line 77: suggestion: rather than configuring them here, I'd put them in `config/config.exs` (and then maybe load them here with `Application.compile_env!`)

rfcs/0000-arrow-user-roles.md
```diff
@@ -0,0 +1,168 @@
...
+and one to work with permissions. This module will centralize all authorization
+questions.
+
+```ex
+defmodule Arrow.Permissions do
+  alias Arrow.Accounts.User
+
+  @required_groups [
+    delete_bar: [:admin],
+    do_foo: [:admin, :viewer],
```
paulswartz, line 76: question: What about `{verb, noun}` instead of a single atom? `{:delete, :bar}`, `{:activate, :foo}`. That also opens up a pattern where you can use pattern matching to allow permissions for a class of verbs or nouns: `{_, :bar}` for all `:bar` actions, for example, or `{:read, _}` for read-only access.

PR #10, opened 2022-01-07
RFC 10: standardize on Colima for local container management

paulswartz:


losvedir:
A thought I just had: if we're worried about docker compatibility, is there a worthwhile approach where we use a Linux VM (e.g. canonical's multipass) and then just use the actual docker inside it?

paulswartz:
@losvedir I found this article about doing that (https://ubuntu.com/blog/replacing-docker-desktop-on-windows-and-mac-with-multipass) but the naive approach didn't provide access to the local filesystem (so `docker build -t <tag> .` didn't work). I'm happy to investigate more if there are other resources you had in mind.

rfcs/0000-docker-desktop-replacement.md
```diff
@@ -0,0 +1,115 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+Podman (like similar tools on OS X) works by running a virtualized Linux VM, and
+then running the containers inside that VM.
+
+To install Podman:
+
+```bash
+brew install podman
+podman machine init --disk-size 20
```
lemald: I also had to run `podman machine start` after this to actually build my image.

lemald:
Okay, well I'm having a bit of trouble actually getting one of my apps properly configured to actually run locally in a container with all of the right environment variables, etc., but overall Podman seems to work for me. It's a bit easier to set up than Minikube and doesn't have all of the confusing Kubernetes baggage that we don't need for our purposes.

lemald:
Update: tried running Concentrate locally at Paul's suggestion (since it requires less fiddling with environment stuff to run than many other apps). The build went fine and the container seems to have started up just fine when I ran `podman run concentrate`.

losvedir:
> I'm happy to investigate more if there are other resources you had in mind.

Nah, `podman` seems to work well! Thanks for the write up. I tried it with a number of our apps (`api`, `prediction_analyzer`, `commuter_rail_ boarding`, `arrow`, and `rtr`) and didn't run into any issues.

I mostly just build locally to quicken the build-test loop when modifying the Dockerfile for, e.g., a new version of erlang, and so this suffices for me!

amaisano:
Re: Lando + Podman:

image

But not currently supported.

It's likely I and other CMS local devs would need to switch to a pure Podman (no affordances and conveniences with Pantheon for pulling db, syncing files, running remote commands, etc) workflow. I'll keep investigating Podman to see if they have any Drupal friendly tips.

amaisano:
Re: Drupal https://gist.github.com/spotzero/18fd821cb9081c666a629e940f166f52

paulswartz:
@amaisano do you know if Minikube's Docker API implementation works with Lando?

paulswartz:
Per the RFC process, I'm calling for a Final Comment Period with a disposition to Accept this RFC.

amaisano:
> @amaisano do you know if Minikube's Docker API implementation works with Lando?

I do not know. I can ask the community.

jzimbel-mbta:
After spending a few hours attempting to use podman instead of docker with VS Code's remote-containers extension, which enables containerized local development, I would say (for this specific use case, on macOS) podman is **not** a simple drop-in replacement for docker.

On macOS, podman does not currently support mounting local directories inside of the container VM, which is a basic requirement for the devcontainer workflow. There is at least one documented workaround, but the VS Code extension does not provide fine-grained enough user configuration to use it.

We might be able to use podman if we switched our host OS to some flavor of Linux, since it supports mounting host directories in that case, but IMO that is too much initial setup as long as we as a department are set on using Macs.

grejdi-mbta:
I'm going to second everything said in the [comment](https://github.com/mbta/technology-docs/pull/10#issuecomment-1017844222) above. Volume mounting is key to having some persistency of data. My main use case is Postgres DB data, but being able to not have to re-build the container every time the files change is also very beneficial.

paulswartz:
Calling off the FCP for now, to address feedback about volumes.

@grejdi-mbta @jzimbel-mbta @amaisano I've updated this document to propose using Colima instead, which for me supported port forwarding and volume mounting without any additional setup. Could you please try it out (with the updated instructions) and let us know if you run into any issues?

rfcs/0010-docker-desktop-replacement.md
```diff
@@ -0,0 +1,117 @@
...
+
+[Colima][colima] provides a Docker-compatible runtime for containers on macOS. It supports port forwarding and volume mounting out of the box.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Colima (like similar tools on OS X) works by running a virtualized Linux VM, and
+then running the containers inside that VM.
+
+To install Podman:
```
grejdi-mbta:
```suggestion
To install Colima:
```

paulswartz:
@jzimbel-mbta @amaisano have you had an opportunity to try out Colima + Docker?

jzimbel-mbta:
Not yet. My team and I have switched back to doing local development without a devcontainer for now. I will try to find time to test out Colima later this week.

grejdi-mbta:
Small update here. I've been running docker through Colima for the past 3 weeks, and it's been fairly OK. Everything seems to run fine, with few exceptions. I have noticed a significant amount of power going to running the containers. I don't have data here, just observation based on my typical usage. Additionally, I've had to sacrifice the state of my database, and mostly just recreate the database through the Dockerfile and using migrations. Will continue to update if I get more info on the second issue, which is more important to development.

paulswartz:
@grejdi-mbta can you say more about your database issues? I've been using it for PostgreSQL databases in multiple applications and haven't had an issue with the data being lost.

grejdi-mbta:
@paulswartz The issues stems from not being able to mount a writable volume on Mac, so database state is lost. There might have been an update lately to address this, but I haven't checked it out yet. Will plan on it.

paulswartz:
@grejdi-mbta do you need to have the database state in a writable volume? I do something like this:
```bash
$ docker run postgres -d -p 5432:5432 -e POSTGRES_PASSWORD=<password> --name <app>_postgres
```

This will run the container in the background, but provide access on the normal Postgres port. To stop the container, but maintain the data:
```bash
docker stop <app>_postgres
```

And then to restart:
```bash
docker start <app>_postgres
```

paulswartz:
This isn't a perfect solution, but I think it's the best suggestion we can make under the current circumstances. I included the writable volume issue out as a drawback

I'm re-calling for the FCP, and will merge at the end of the week if there are no blocking comments.

grejdi-mbta:
[non-blocking] I was a having trouble getting Spark running inside a container built with Colima, but no issues with Docker Desktop. I was not able to investigate further due to time constraint.

PR #11, opened 2022-01-26
rfc: update example message in CloudEvent

digitalcora:
Although there is no spec for the `data` format itself, this example could be misleading, since we have removed the `received_time` field from the real events.


losvedir (approved):
👍

PR #12, opened 2022-03-10
rfc: CentralStation aka Superloc

losvedir:
[Rendered](https://github.com/mbta/technology-docs/blob/5ea17dd79e50b7e5bb366059aae33d83e828a480/rfcs/0000-central-station.md).


rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+CentralStation is the app that centralizes all the realtime vehicle data and is the canonical
+source of that data for CTD apps and our public feeds. Data is at the core of CTD, and where are
+vehicles are and what they're doing underlies all of that. It deserves to be centralized, easily
+accessed and queried, and valid and referentially sound by construction. The question of what trip
+a vehicle is on is rightly the domain of the MBTA, and shouldn't be left to Swiftly, a 3rd party,
```
paulswartz, line 52: I get where you're coming from: really all the real-time data is the domain of the MBTA, not only the vehicle locations. We pay Swiftly because they do a good job of making predictions and assigning trips to vehicles without much input from us. If we stop using that functionality we're effectively paying for it twice, or worse: we're paying for it twice and not doing as good a job.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+predictions (of unknown quality, or perhaps simply the arrival/departure values from the
+underlying trip + offset), and these are published in a TripUpdates file which concentrate
+consumes directly.
+
+Conversely, "block waivers", when trips are canceled, are also handled by Busloc and also bypass
+Swiftly. Busloc receives the list of waivers via TransitMaster and publishes them as a special
+TripUpdates file, marking their stops as "SKIPPED". Swiftly is meanwhile publishing predictions
+for the canceled trips because the VehiclePositions are in the feed, and Swiftly can't handle
+waivers. So concentrate merges the Swiftly TripUpdates and the special Busloc TripUpdates feed and
+cancels the trips there. (Is that correct?)
```
lemald, line 234: Note that this information is also consumed by Skate to generate notifications (we take StopTimeUpdates and roll them up into individual notifications as appropriate).

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+concentrate, by selecting its revenue vehicles and public fields. In the future, it may produce
+the Alerts and TripUpdates fields, too, using Swiftly and/or RTR and/or TransitClock as
+implementation details, and applying alerts and other business logic on top. But for now, we lean
+on the existing alert-applying logic in concentrate, and expect RTR and Swiftly to provide the
+TripUpdates to concentrate.
+
+If internal CTD apps need data not in the public feed, they can access it via CentralStation's
+API. For example, Glides will use CentralStation to access all the non-revenue vehicles, and Skate
+can get block and run information. (And operator info, like name? I don't know if PI info like
+that needs to live in CentralStation.)
```
paulswartz, line 105: If the operator information doesn't live in CentralStation, and there isn't a Busloc GTFS-RT feed, where does Skate get that information?

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+It will consume VehiclePositions data from CentralStation, and importantly, those vehicle
+positions will have a GTFS trip tagged. This will simplify RTR as a large part of the message
+handling logic is about deciding what trip a vehicle is on.
+
+It will also consume the ServiceChanges feed, which is how CentralStation will convey the TSCH
+offsets it receives.
+
+As a bonus, by relying on GTFS trips and ServiceChanges, RTR won't have to rely on handcrafted
+patterns, and should be able to predict, e.g., new stations or closed stations automatically.
```
paulswartz, line 196: I don't see how this follows: there's no new information in CentralStation, so how would RTR be able to predict new stations?

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+those extra fields above. Glides consumes it directly, and would ideally like the VehiclePositions
+of all light rail trains, in service or not.
+
+### With CentralStation
+
+RTR will change to behave more like how we want Swiftly to behave.
+
+It will consume VehiclePositions data from CentralStation, and importantly, those vehicle
+positions will have a GTFS trip tagged. This will simplify RTR as a large part of the message
+handling logic is about deciding what trip a vehicle is on.
```
paulswartz, line 190: It's also deciding what stop a vehicle is at, which seems like it would need to bring in much of the Pattern functionality from RTR.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+Busloc primarily keeps track of the whereabouts of MBTA buses and publishes a few feeds related to
+them. Busloc runs on the MBTA network because one of the sources of its data, the TransitMaster
+DB, is only accessible from there.
+
+The majority of what Busloc deals with is "normal" bus data. These are buses logged on and running
+their expected, planned routes. Busloc gets this data from TransitMaster, via its hacky internal
+spelunking of the TransitMaster DB, and via Samsara, a cloud service that has more frequent GPS
+updates. Busloc combines these feeds, taking the latest lat/long from either source for every
+vehicle. These logged-on vehicles have their GTFS trip directly in the data, since TransitMaster
+and GTFS both get their source data from HASTUS. The Samsara data doesn't _have_ to be queried
```
paulswartz, line 214: This is only partially true. Properly logged-in vehicles with a functioning TransitMaster unit will have a correct GTFS trip ID. 

However, there are other cases which Swiftly compensates for by assigning a trip ID:
- broken TransitMaster unit, so only Samsara is providing a vehicle location
- Operator logged on to a different piece of work, so the assigned GTFS trip ID does not match what they're actually doing
- Winthrop, which I don't think gets the same schedule data (@dks-mbta would be able to say more)

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+cancels the trips there. (Is that correct?)
+
+Lastly, Busloc has another Samsara feed from Yankee, which includes the VehiclePositions of
+shuttle buses. Those are output more or less as-is, and we don't try to make predictions for them.
+(Do these end up in concentrate? I didn't see them. I can't tell where
+VehiclePositions_enhanced_shuttles.json goes.)
+
+In addition to the standard VehiclePositions and TripUpdates data, Busloc publishes operator data
+(e.g. the name of the person driving the bus) from TransitMaster, for Skate to consume. I believe
+this data is passed along in one of the TripUpdates files (?), and stripped out later in the
```
lemald, line 243: Operator info comes through in `VehiclePositions_enhanced.json`.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+WMATA is an example of a transit agency that has a centralized web-app powering a lot of its real
+time information. In a demo given to Jessie, Eddie, and me about single tracking, they pulled up a
+web page which showed the realtime position of various trains, as well as the state of switches.
+In the demo they used the web interface to manually indicate that a switch had changed, indicating
+single tracking, and showed how that affected predictions and signs downstream.
+
+Regarding the use of a DB here, in my past experience I've really appreciated postgres, and the
+data we work with is relational and fits naturally. However, in our experience at CTD with
+prediction analyzer, we can run into performance issues if we're not careful.
```
boringcactus, line 467: Could performance issues be mitigated by running separate instances of CentralStation per mode? It seems like the benefits of consolidating all modes together are mostly at the development level, and keeping modes separate at the operational level would mean performance bottlenecks are isolated to a single mode.

rfcs/0000-central-station.md
paulswartz, line 234: This sounds correct.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+Conversely, "block waivers", when trips are canceled, are also handled by Busloc and also bypass
+Swiftly. Busloc receives the list of waivers via TransitMaster and publishes them as a special
+TripUpdates file, marking their stops as "SKIPPED". Swiftly is meanwhile publishing predictions
+for the canceled trips because the VehiclePositions are in the feed, and Swiftly can't handle
+waivers. So concentrate merges the Swiftly TripUpdates and the special Busloc TripUpdates feed and
+cancels the trips there. (Is that correct?)
+
+Lastly, Busloc has another Samsara feed from Yankee, which includes the VehiclePositions of
+shuttle buses. Those are output more or less as-is, and we don't try to make predictions for them.
+(Do these end up in concentrate? I didn't see them. I can't tell where
```
paulswartz, line 238: The Yankee shuttles are incorporated into the main VehiclePositions_enhanced.json feed. The `_shuttles` feed isn't currently used, and we don't have any buses using that AVL vendor anymore.

(internal CTD users can see more in https://github.com/mbta/busloc/blob/master/config/prod.exs)

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+VehiclePositions_enhanced_shuttles.json goes.)
+
+In addition to the standard VehiclePositions and TripUpdates data, Busloc publishes operator data
+(e.g. the name of the person driving the bus) from TransitMaster, for Skate to consume. I believe
+this data is passed along in one of the TripUpdates files (?), and stripped out later in the
+process by concentrate.
```
paulswartz, line 244: It's in the VehiclePositions_enhanced.json feed. Concentrate ignores fields it doesn't know about, including the operator information.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+TrainLoc and CommuterRailBoarding are two applications that once were separate but were
+consolidated into a single umbrella application, using the name Commuter Rail Boarding, and they
+manage the realtime data of commuter rail trains from Keolis.
+
+Keolis offers two firebase feeds, consumed by CRB:
+
+- The "boarding status" feed, which we get for trains at a few stations, with info like "now
+  boarding on track 9". We put these in TripUpdates_enhanced.json, in our extension
+  `boarding_status` field. Concentrate generally merges these with predictions for the trains, but
+  we often end up with StopTimeUpdates with null arrivals and departures.
```
paulswartz, line 267: This happens when there's a boarding status update for a vehicle (such as "On time") but no actual time information from Keolis. It would be great to fix, but if it was possible we could fix it in CommuterRailBoarding as well.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+specific, though I'm not aware of any other transit agency using it.
+
+In addition to simply merging feeds from different modes, it has logic to process the data by
+doing things like dropping vehicles that don't have a trip ID, or merging in alert data to update
+trip predictions appropriately.
+
+### With CentralStation
+
+Nothing will change immediately. However, more of its logic may move into CentralStation, like
+canceling trips. Longer term, we may even move alerts and predictions into CentralStation,
```
paulswartz, line 326: If cancelling trips moves into CentralStation, how does that information get out of CentralStation and into the API/downstream?

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+obviating concentrate entirely.
+
+## Alerts-UI
+
+### Current description
+
+Alerts-UI is the interface that PIOs, bus dispatchers, and Keolis use to manage alerts in the
+system. It has an API that gets cached by our CDN, and the CDN endpoint is what is polled by
+concentrate, APIv3, and others.
```
paulswartz, line 335: We do query it through the CDN, but it's not an API: it's a file hosted in S3 like the other GTFS-rt feeds.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+and performant.
+
+The status quo works for now, and there's a path forward from here to accommodate Service Changes,
+as detailed in the alternatives section.
+
+While conceptually it makes sense to think of "realtime data" as a single coherent concept, the
+adapters, and the extra data, and the business logic, do vary quite a bit by mode. I expect,
+though, that having it in one place will make it easier to work with. If it's in a DB, then the
+centralized schema will allow the apps to reuse models, datatypes, queries, and helper functions.
+Some commonalities like serializing into GTFS-rt or enhanced JSON files will only need to be
```
paulswartz, line 396: There's already this possibility (for example, using the models that Concentrate or the API expose, or another set in a library). In practice, each application has needed slightly different data and so it hasn't been useful to share the models or GTFS-RT generation code (except to the extent that applications rely on `gpb` to parse/create GTFS-RT output).

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+the MBTA to view and work with that data.
+
+And by keeping all the data in one application, visualization and debugging tools - which, as this
+data being the core of CTD, we should invest in - will naturally apply across all modes.
+
+Lastly, this is a response to staffing constraints. It is easier to keep one application, even a
+sprawling application that does many things, up to date in terms of CI, dependencies, and
+infrastructure. The more we can move day-to-day development into one application, the easier it is
+for developers to maintain a working knowledge of the codebase, and the easier it is to onboard
+new developers. Using a DB as the primary datastore should be easier for new developers, since
```
paulswartz, line 417: I'm not sure how true this is. The vehicle location sources that you've mentioned are very different, as are the ways that staff will interact with the data. Will working on the Trike/Kinesis ingestion transfer much to the CRB Firebase endpoint?

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+The biggest alternative is to build on the status quo. Currently, all the modes work tolerably, if
+separately. Service Changes can be accommodated by Skate, Glides, and Keolis separately. We can
+lean on Swiftly for vehicle tagging.
+
+One downside of this is it locks us much more tightly into Swiftly. If we rely on Swiftly for
+tagging, and have several apps all using Swiftly's API, then we are tightly coupled to them,
+making it hard to move off them in the future (if our TransitClock experiments prove fruitful), or
+giving them leverage to raise prices quite a bit (like Google did with Maps). If Swiftly does a
+poor job of tagging vehicles, we have no recourse.
```
paulswartz, line 430: Wouldn't we have the same recourse we have now, to write our own application?

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+In the demo they used the web interface to manually indicate that a switch had changed, indicating
+single tracking, and showed how that affected predictions and signs downstream.
+
+Regarding the use of a DB here, in my past experience I've really appreciated postgres, and the
+data we work with is relational and fits naturally. However, in our experience at CTD with
+prediction analyzer, we can run into performance issues if we're not careful.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
```
paulswartz, line 471: A big question for me is how does this perform/scale in practice? Right now, Concentrate can process all the GTFS-RT pretty quickly (every couple seconds) because it's all in-memory and doesn't need to write to a database. Additionally, the logic for processing the GTFS-RT can be simple because the other applications generate valid inputs.

rfcs/0000-central-station.md
digitalcora: Service changes often cut across modes (e.g. a replacement of part of a subway line with shuttle buses). If there were separate data stores for each mode, they would need ways to communicate amongst themselves and keep their data consistent, somewhat undoing the benefits of everything living under one roof.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+that will all need to switch from the API to ServiceChanges approach.
+
+Developers will continue to have to maintain working knowledge of, and actively build on and
+debug, a variety of similar yet different applications.
+
+Another alternative is a scaled-down CentralStation. It could be the centralized interface for
+Glides, Skate, and Keolis to post Service Changes, but not maintain the state of vehicles itself.
+This helps somewhat with the Swiftly coupling, but without having the vehicle and trip state, it
+can't validate the requests, or perhaps easily know when one has expired. It also doesn't support
+vehicle tagging, which is a major part of what these tools will do.
```
paulswartz, line 443: Why couldn't it support vehicle tagging? If it's generating a feed for Swiftly to consume, can't it include a VehiclePosition with a trip assignment?

rfcs/0000-central-station.md
boringcactus: Ah, yeah, if changes frequently don't apply to modes independently then it'd probably be unhelpful to silo data by mode for performance reasons.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+![Proposed setup](imgs/realtime_proposal.png)
+
+# Motivation
+
+[motivation]: #motivation
+
+As was brought to light in the TRC team structure iteration plan, realtime data has until now been
+scattered across various apps, and passed around via standard and non-standard GTFS-esque files.
+With a renewed focus on the realtime data, and one team to manage it, it makes sense to
+consolidate the data into one app as well.
```
paulswartz, line 34: This is how things used to work: MBTA-Realtime consolidated all the real-time data into one database. and provided the GTFS-RT output as well as an API for clients. We moved away from this model because the different modes had different needs, and having all the data go through a single database-backed application didn't perform well.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+Dashboard, and be the API that Keolis, Glides, and Skate uses for Service Changes. It will be the
+canonical source of all realtime data that CTD deals with, public and private, will publish the
+standard, public data as GTFS-rt and provide API's to internal apps for manipulating data or
+retrieving non-public and non-standard data.
+
+Current realtime data ecosystem:
+
+![Current setup](imgs/realtime_current.png)
+
+Proposed realtime data ecosystem:
```
paulswartz, line 23: This doesn't have any files in it, and it also doesn't break out the separate Swiftly environments. I think that's part of what makes it looks so much simpler.

rfcs/0000-central-station.md
dks-mbta: Operator data is also made available in the Swiftly dashboard. Swiftly Playback is a very useful tool, and some Bus Operations folks use it.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+## Service Changes
+
+A major component of this proposal, and a strong focus of CTD going forward, is the idea of
+"Service Changes". This is a [GTFS-rt proposal that has been discussed for 3.5
+years](https://github.com/google/transit/issues/113). Here is the [most recent version of the
+proposal](https://docs.google.com/document/d/1oPfQ6Xvui0g3xiy1pNiyu5cWFLhvrTVyvYsMVKGyGWM/edit#heading=h.mm6bblup6h8d).
+I refer to these in the RFC as a "ServiceChanges" file, but really the proposal is about new
+FeedEntity enum values, as well as extending the existing TripUpdates one. By convention, we
+produce `VehiclePositions` and `TripUpdates` files separately, though this is not required by the
+spec. We could similarly produce a `ServiceChanges` file composed of feed entities related to
```
paulswartz, line 124: Yes, I think a separate feed is the way to go here. The way I think about it is how frequently the data is updated:
- VehiclePositions: every second
- TripUpdates: every 10 - 30 seconds
- ServiceChanges: every 1 - 15 minutes

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+Swiftly has implemented some of it, in spirit, via its [Service
+Adjustments](https://www.goswift.ly/blog/introducing-service-adjustments) feature. Swiftly
+currently only supports creating these via a proprietary
+[API](https://swiftly-inc.stoplight.io/docs/service-adjustments/b3A6Mjg0MzkyMzY-create-adjustment)).
+
+The main thrust of it all is a way to update GTFS trips in real-time more comprehensively. Current
+TripUpdates can make predictions against existing trips, and can indicate that a train will skip a
+stop, but it can't indicate that the trip will go to a different stop, or provide a shape for that
+new route, or update properties of the trip like its headsign, limiting our ability to provide
```
paulswartz, line 135: This sets out a large chunk of work for CentralStation, in additional to all the other features: drawing new shapes is a complicated interface!

paulswartz (commented):
Things I love in here:
- reducing the scope of Busloc, at least the part which has to live inside the MBTA firewall
- providing an API to Skate/Glides/Keolis for internal information (both reading and writing)
- thinking about how to integrate ServiceChanges and tagging in a generic way

Some other things to think about:
- Skate needs an interface to write information back into TransitMaster (potentially including vehicle -> work assignments). Do you think that interface would be to CentralStation instead, and CentralStation would talk to TransitMaster?
- How does the information which needs to go to the signs & PA/ESS fit in?
- How does Lamp (which will want to consume the raw data as much as possible) fit in?

I'm less sure that pulling everything into a single app is going to make things simpler to understand. For example, "combining similar things" is why CommuterRailBoarding and TrainLoc were combined, and it doesn't sound like it's made them easier to understand or work with. 

It's a complicated problem, with a lot of moving pieces!

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+Busloc primarily keeps track of the whereabouts of MBTA buses and publishes a few feeds related to
+them. Busloc runs on the MBTA network because one of the sources of its data, the TransitMaster
+DB, is only accessible from there.
+
+The majority of what Busloc deals with is "normal" bus data. These are buses logged on and running
+their expected, planned routes. Busloc gets this data from TransitMaster, via its hacky internal
+spelunking of the TransitMaster DB, and via Samsara, a cloud service that has more frequent GPS
+updates. Busloc combines these feeds, taking the latest lat/long from either source for every
+vehicle. These logged-on vehicles have their GTFS trip directly in the data, since TransitMaster
```
lemald, line 213: One thing to keep in mind is that whenever a bus has a GTFS trip ID, that's because something else (TM, Swiftly, etc.) has already done its own trip matching logic. Operators only log in with run ID, so if you get trip ID from something else you're already relying on some other service's algorithm.

rfcs/0000-central-station.md
dks-mbta: Yeah, I'd think RTR needs to have patterns at the track circuit level, which is not available from GTFS.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+
+### With CentralStation
+
+Skate will get all its vehicle position data from public feeds. It will supplement it with private
+data from CentralStation.
+
+Depending on if we want CentralStation to store PI operator data, it may continue to get that
+information from an S3 file produced by Busloc.
+
+It will use CentralStation's API for train tagging and service changes.
```
lemald, line 310: I assume you mean bus tagging? And in this future world, are we imagining that associating a bus to a trip would still be roughly as automatic as with Swiftly, just with a manual override?

rfcs/0000-central-station.md
dks-mbta: We use Swiftly's trip ID primarily. The TM one is only used in the case of overloads. We decided Swiftly's trip assignments are better (though it's an open question). Swiftly also makes use of modifications in GTFS that TransitMaster doesn't know about (mid-rating stop and route changes, though-routed school trips that are split into multiple trips in GTFS, etc).

We don't have any live TM data for the Winthrop routes (712 and 713). It's purely Samsara and Swiftly.

rfcs/0000-central-station.md
dks-mbta: I'll add that bus trip assignment is a hard problem, so that would be a major project in itself.

rfcs/imgs/realtime_current.dot
```diff
@@ -0,0 +1,88 @@
...
+  "Swiftly (mbta-winthrop)" -> "https://api.goswift.ly/real-time/mbta-winthrop/gtfs-rt-trip-updates";
+
+  "s3://mbta-gtfs-s3/rtr/TripUpdates_enhanced.json" -> "Realtime Signs";
+  "s3://mbta-gtfs-s3/rtr/VehiclePositions_enhanced.json" -> "Realtime Signs";
+
+  "s3://mbta-gtfs-s3/VehiclePositions.pb" -> "Glides";
+  "s3://mbta-gtfs-s3/TripUpdates.pb" -> "Glides";
+
+  "https://api.goswift.ly/real-time/mbta-bus/vehicles" -> "Skate";
+  "s3://mbta-busloc-s3/prod/TripUpdates_enhanced.json" -> "Skate";
```
skyqrose, line 81: **nitpick:** Skate also uses the API, to show train locations on the shuttle map.

rfcs/imgs/realtime_proposal.dot
```diff
@@ -0,0 +1,39 @@
+digraph G {
```
skyqrose, line 1: **meta feedback about the RFC:** These two graphs are at different levels of abstraction, so seeing that this graph is less complex than the other doesn't really say anything. I think it would be easier to see this proposal's effect on CTD's architecture if this graph were as detailed as the earlier one.

It would also be nice if this graph reflected the desired end state (with concentrate completely absorbed, with Service changes, etc). Maybe also with a separate phase 1 stepping stone graph if that's important to the proposal.

I'd also like to see detail about the data flow within the app. e.g. how raw data becomes trip assigments becomes vehicle positions becomes predictions. Just because it's within one app doesn't mean we can ignore the complexity of that pipeline.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+- Start Date: 2022-03-07
+- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
+- Asana task: [asana link](https://app.asana.com/)
+- Status: Proposed
+
+# Summary
+
+[summary]: #summary
+
+"CentralStation" (aka "Superloc", but official name TBD) is an application that manages all of the
```
skyqrose, line 11: **Name suggestion: "Union Station".** (It joins the different parts of the realtime data flow into one, just like Union Stations join different railroad companies into one station.)

rfcs/0000-central-station.md
skyqrose: They don't do a perfect job though, for example with block overloads, and if we put in the effort we might be able to do a better job with all of our internal knowledge. But that's a separate issue from unifying how we handle all the parts of the realtime data flow. If we don't want to redo trip assignments, can we meaningfully do Central Station and allow Swiftly to keep doing the trip assignments?

rfcs/0000-central-station.md
skyqrose: Operator info should definitely live here. This seems like it should be The Place for internal information like that. If Glides/Skate have to look elsewhere, then it's not really meeting its goal of being a one-stop-shop / single source of truth.

rfcs/0000-central-station.md
skyqrose: I imagine a separate app for creating Service Changes, and then CentralStation consumes that standard format to publish them / forward them to swiftly / adjust vehicle positions or whatever needs to happen with the data.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+vehicle. These logged-on vehicles have their GTFS trip directly in the data, since TransitMaster
+and GTFS both get their source data from HASTUS. The Samsara data doesn't _have_ to be queried
+from the MBTA network, but is done to have all the data in one place and for simplicity merging
+with the TransitMaster data. This data is published as VehiclePositions, which is consumed by
+Swiftly to generate TripUpdates. The data is GTFS-rt-esque, but includes block and run data, which
+Swiftly takes advantage of.
+
+In addition to the planned service, TransitMaster also has "overload" trips, which is when a bus
+is assigned to run a route that it wasn't planned for. These create "ADDED" trips, based on a
+scheduled trip but with an offset at every location. Since Swiftly cannot produce predictions for
```
skyqrose, line 222: I know Swiftly can't do trip assignments for block overload / ADDED trips, but if we did our own trip assignments, would Swiftly be able to handle making predictions for them?

Related: If we did our own trip assignments for shuttles would Swiftly be able to make predictions for shuttles?

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+### Current description
+
+Right now Glides is just a consumer of the realtime data, directly from RTR.
+
+### With CentralStation
+
+Glides will either get all its data from CentralStation, or get the public data from our public
+feeds, and supplement it with private data from CentralStation.
+
+It will use CentralStation's API for train tagging and service changes.
```
skyqrose, line 293: It would be nice for Glides to get everything from one place with a single API.

The non-public data we want out from our realtime data is any not-in-service vehicles that would be filtered out of the public feed, and the raw (not geofenced) GPS data (which would need to be an extra field, not part of the normal latlng field.)

That GPS data we want to be as low-latency as possible. What will the delay be like getting it from Central Station? Will we need to get it from another source anyway just to keep the delay lower?

rfcs/0000-central-station.md
skyqrose: The same way that data currently gets from concentrate to the API. Central Station can publish public GTFS-RT feeds to s3, and the API can read them.

rfcs/0000-central-station.md
skyqrose: What information would Central Station need to be remember over time in a database, instead of in memory? I guess whatever history is needed for trip assignments, as well as service changes. But I don't see why the flow of publishing new latlng's trip assignments needs to wait for database writes.

Also, presumably any feature that requires slow database writes or GTFS-RT edits will need to happen regardless of whether it's in a single app like Central Station or spread between a bunch of smaller apps.

I would think that centralizing everything in one app would decrease latency because it cuts down on network calls between apps.

rfcs/0000-central-station.md
```diff
@@ -0,0 +1,507 @@
...
+## Skate
+
+### Current description
+
+Skate consumes bus realtime data, straight from Busloc. This data is GTFS-rt-esque, enriched with
+operator information and block and run data.
+
+### With CentralStation
+
+Skate will get all its vehicle position data from public feeds. It will supplement it with private
```
skyqrose, line 304: **overall thoughts about being a single source of truth:**

Yes! I love this proposal. I have been frustrated by the lack of a single source of truth for internal data for a while. I think Central Station should lean into that, with the goal of Glides/Skate not needing to ask any other source for realtime data. So publishing an internal extended feed, instead of just a supplement. That way, they don't have to deal with merging and matching data, or with data from different sources being slightly out of sync. My dream is for Central Station to publish a "public_gtfs_rt", a "glides_gtfs_rt_with_light_rail_and_some_internal_data.json", and a "skate_gtfs_rt_with_bus_and_other_internal_data.json".

I think that's best if Central Station plans to envelop all of concentrate and alerts, and become the final aggregator/publisher of data.

skyqrose: (Side note: a single source of truth for schedule data (combining GTFS+HASTUS) might be even more important than realtime data, but I think it's okay if schedule data stays separate from realtime data, and a schedule SSOT would be a separate project.)

lemald: > (Side note: a single source of truth for schedule data (combining GTFS+HASTUS) might be even more important than realtime data, but I think it's okay if schedule data stays separate from realtime data, and a schedule SSOT would be a separate project.)

For what it's worth, my hope is that the ODS project might provide the first step towards accomplishing this, though it won't be simple. The code to squish together the GTFS and HASTUS schedules is definitely one of the more complicated parts of Skate, not to mention the startup overhead it introduces.

rfcs/0000-central-station.md
lemald: @skyqrose is there any chance that Glides users might want a track circuit level view of the Green Line (for the areas where that exists, at least?). That seems like it might be relevant as well.

rfcs/0000-central-station.md
digitalcora: One of the meta-level thoughts I had on this RFC is that it says surprisingly little about how "static" data fits into this picture, even as we are looking to introduce features that involve more extensive integration of "static" with real-time data. Contra @skyqrose I might even propose that a single-source-of-truth for _all_ data, whether it is deemed "real-time" or not by the GTFS standard, would be worthwhile to consider. But I think making this a manageable amount of work would also require a fundamentally different approach, and I have enough thoughts here that it could honestly be its own competing RFC.

rfcs/0000-central-station.md
lemald: Yeah, having worked on both RTR and the bus tech stack, I can say that they're fairly different, and not just in terms of superficial details like data format.

rfcs/0000-central-station.md
lemald: Even if we do build such an app, as part of CentralStation, whose job is it to input such information into it? Dispatchers? Inspectors? PIOs? How does it fit into their existing job duties? Would we get buy-in from leadership in the relevant parts of ops to have their employees putting work towards this? Some of this is for stuff that already gets put into an alert or something, but this is introducing a new level of detail and complexity.

rfcs/0000-central-station.md
skyqrose: Yes definitely. And though the track layout is static/schedule data, the assignment of trains to those track circuits in real time definitely belongs to this RFC, and would be great to have in any internal realtime data feed.

rfcs/0000-central-station.md
digitalcora: > having all the data go through a single database-backed application didn't perform well

Since we have been talking about databases a good bit on TRC, and definitely want to avoid repeating the mistakes of history, I'm curious how this system was implemented and what the performance obstacles were. Is there old documentation on it anywhere?

losvedir:
Thanks everyone for your comments! I have been working and re-working how I want to incorporate it and respond to it, but in the interest of forward progress I'm posting what I have now.

I think this PR has conflated several different ideas I have about realtime data, and it would be good to tease them apart. They are:

- Centralization of realtime data
- Trip ID tagging (and "current stop" and "stop status" determination)
- Support for Service Changes
- More dev tooling and data introspection
- Updating RTR to be less manual

## Centralization of realtime data

This is about having one app that is the canonical source of data. There are bits of data produced by the mode-adapters: RTR, Busloc, and CRB; plus Swiftly; plus Concentrate. Our other internal apps currently consume and merge data from multiple sources, a tedious and error prone process.

This part of the proposal sets CentralStation as the interface for all our internal apps to get their data, as well as the public apps. For example, it would keep information like bus operator name and non-revenue vehicles, and it would output different feeds - a glides.json with the all the light rail data including non-revenue vehicles, a skate.json with all the bus vehicles including operator data, and the public data feeds, in the format of GTFS-rt.

@skyqrose especially liked this aspect of it. @skyqrose - can you provide an example of some issues you've seen with the current realtime data ecosystem? I believe you gave me an example something about matching up data between two sources which were 6 seconds off, and having missing elements in one or the other?

There is some complexity in merging of the data, though it's better to be done here than in all the client apps. For example, BusLoc produces extra data like operator information, which we would want to intake. However, that vehicle also goes to Swiftly, and Swiftly could potentially modify the vehicle information (e.g. the trip the vehicle is on, or the stop it's approaching), while dropping the operator information. Concentrate can further change the vehicle, for example by rounding the lat/long and bearing, or even dropping the vehicle altogether if it lacks a trip.

How would we handle these scenarios?

- The bus operator for a vehicle is set to NULL in the BusLoc data; a naive "merge all feeds, overwriting fields with newer data, but leaving unknown fields alone" might not pick up the loss of operator. (This is the algorithm concentrate uses, and I'm not sure concentrate's behavior here? Maybe we don't ever zero out operator name.) We would likely need to explicitly handle and prioritize data sources and fields from each.
- What if some of concentrate's filters should apply to some of the outputs but not the others? The public feed would benefit from concentrate removing non-revenue vehicles but not Glides.

The benefit of this approach by itself is that it helps internal app developers have a unified source of data, and makes it clear to them whom to request data from and how they would receive it.

### Can concentrate be this?

@paulswartz and I discussed updating Concentrate to be this app. Right now, it *already* merges all our available realtime data. Architecturally it consumes a number of data feeds in a variety of formats, "concentrating" it into its in-memory data store. It then serializes it periodically into a number of formats (pb, json). We could introduce a new concept of an output adapter, and make the filters and encoders run differently per adapter. We could have the same centralized store of vehicles and trip updates, and for the public feed run them through all the current filters, and then serialize into GTFS-rt JSON/PB files, but also run them through the "glides" adapter which would omit the "drop vehicle if non-revenue" filter, and serialize into JSON only, and the "skate" adapter could serialize into JSON, specifying it wants the operator info fields.

I think that this part of the proposal _by itself_ makes sense. However, if we want to explore the other parts, it might not work as well.

Trip tagging _could_ potentially work, as a very complicated step of the Concentrate pipeline.

Service changes don't fit in very well. Or, at least, we could shoehorn it in probably as well as we could into any existing app.

Developer tooling and introspection could be somewhat implemented. The data model isn't really conducive the relational DB aspect, but the performance implications of that still need to be tested. We could provide some light checking in the various "producers" to check for staleness and log warnings there, and we could extend the "pipeline" concept to have a "validation" step which alerts for missing predictions. Concentrate already consumes GTFS to some extent, but we'd need to update it to store more static data depending on the complexity of the validations we want. Still, the data is stored internally as one giant table of all data, making it maybe a little less appealing to work with, complicating ways we could display dashboards of it and otherwise introspecting it.

## Trip tagging

This is the functionality that takes a given vehicle and says what GTFS trip it's on, if any. It tries to match against the public static GTFS feed and could potentially create ADDED trips if there are none. In addition it calculates the vehicle's current stop_id, stop_sequence, and current_status.

There was a lot of pushback on this aspect and its difficulty, especially given that we're paying Swiftly for it.

RTR already does this to some extent. Any investments we made in that business logic would have positive effects if we were doing this for all the modes. However, RTR currently doesn't require the tagging to be very accurate, so there is less pressure to _make_ it that accurate, in contrast to modes that use Swiftly for predictions.

Tagging already happens in a number of places: TransitMaster for buses takes a first stab, with updates from Swiftly. Keolis's AVL system also does most of the time, again with updates from Swiftly. Some Keolis issues observed:

- Trains that stop at stations with the control car far enough away from the geofence don't get recorded as "arrived" and end up being very "delayed" when sitting at the terminal.

Trip tagging could be done better with run and block information. Is this generally available? RTR does not take advantage of this at all. We pass run information to Swiftly in CR but I don't know if they use it. There's the converse problem of trips beginning at major terminals like South Station maybe _not_ following the run/block, if a previous train is delayed.

## One interface for service changes

One of the motivating factors prompting this RFC, and a reason to do it now rather than later, is to get ahead of "service changes" work that will presumably coming up. Keolis, Skate, and Glides are all talking about ways they may want to allow people "in the field" to update realtime data.

However, the upcoming work is pretty varied, and maybe it makes sense to let each team handle it on their own, independently via the Swiftly API if need be. It would be good to get some concrete thoughts by the Skate and Glides teams, answering:

1. What, specifically, do you anticipate your users will want to be able to modify, and maybe a rough timeframe?
2. Absent any action on our part, do you have thoughts on how you'd accomplish that?

According to a document shared with us from Keolis, this is what they have:

1. Near term: update trip-vehicle assignments, in the case where their AVL system is not fully functional and Swiftly's auto-assignment is incorrect. Long term: Implement a "schedule editor" that allows at least express-to-local conversion (added stops), local-to-express conversion (removed stops; re-routes), changes to flagstop status, unscheduled (extra; rescue) services, cancellations/terminations.
2. They would integrate with Swiftly directly, hitting the Swiftly API from their internal software.

What are the thoughts here for Glides? It will almost certainly require changes on the realtime data side; RTR currently has no way to consume updates (well, short of Glides putting OCS messages in Kinesis, I suppose!). How much will folks in the field want to update Glides _purely_ for predictions? For example, saying "this train will leave in 6 minutes" is not that useful, and somewhat annoying, if it doesn't also help the trains leave in 6 minutes in some fashion. In other words, aren't predictions somewhat lower priority to Glides users than operational concerns?

What are the thoughts for Skate? I've heard of plans to integrate with the Swiftly ServiceAdjustments API several months down the line, but I'm not sure to what extent. For example, it's probably not feasible to expect a Skate user to pick a particular trip by ID that a vehicle should be on, when there are so many of them on bus (contra CR where trips are named and known). And thoughts on how this data would get entered and into the data ecosystem? TransitMaster and then extracted from BusLoc or directly in Skate? Maybe limited to just passing along block waivers or something to that effect?

## Developer tooling and introspection

The RFC called out using a relational DB to manage the realtime data and also having the app run as a Phoenix webapp. These are implementation details and could be potentially layered onto any aspect of the other proposals accepted.

### Relational database

GTFS data is fundamentally relational. None of our current apps exploit this fact by using a relational DB, though earlier versions of the API (v2) and gtfs_creator (v1) used databases to some extent. There were performance problems, but I don't know exactly why or how the databases were used.

The performance is an open question and we'd want to figure out expected read/write rates and do some experiments.

One of the main benefits of having the data in a DB would be to take advantage of Ecto, which would make it easier to, e.g., query Red Line vehicles joined with their predictions.

### Introspection via Phoenix webapp

There are a variety of ways that we introspect our current tools.

- [Realtime Dashboard](http://realtime-dashboard.mbtace.com/pb/rtr-prod) puts the current VehiclePositions and TripUpdates on a line diagram, and can select among all the environments of RTR and Concentrate. It also has rudimentary support for displaying RTR's internal state, with links to "Train Sheets", "Vehicles State", etc, based on the backing S3 files, since RTR persists its state to S3 in case it needs to restart.
- Splunk alerts monitor logs from a lot of the applications for various triggers, e.g.
  - RTR alerts if there are not enough TMOVs in a minute
- API Checker alerts on the state of a number of apps by "proxy" of the public feed:
  - If bus/CR/subway predictions are stale/empty
  - Limitations
    - when things go wrong, frequently multitude of alerts
    - difficulty matching up with GTFS and other more sophisticated alerting
    - when something is "stale" - where in the pipeline did it go stale?
- Devops runbooks typically have an ["identify the cause" section](https://github.com/mbta/wiki/blob/master/devops/runbook/CR_Data_Outage.md#identify-the-cause), which has the person query a variety of feeds and look at different values in the feeds to see where in the pipeline an issue began. For example, in this linked runbook:
  - Splunk logs for TrainLoc, to see if any Keolis errors appear
  - TrainLoc's S3 output to make sure the header's timestamp is current
  - Check concentrate logs
  - Logging into Swiftly dashboard
  - Check the API Dashboard in case it's an issue with the API

For the introspection, I'm imagining an admin dashboard with timestamps of last updates from a variety of sources, counts of predictions, maybe compared to current scheduled trips. There could potentially be vehicles plotted on a map, and trip updates on a ladder diagram. For a given vehicle, maybe we could see its data through the system from, e.g., Keolis -> CRB -> Swiftly -> Concentrate.

### Control via the webapp?

This works better if everything is centralized in one app. In that case, we could have certain "superuser" powers to work with the data.

[Logan Nash in Slack](https://mbta.slack.com/archives/C01B0577NH4/p1647021697216739):

> Idea: a “kill switch” for rider-facing predictions in cases like this where the data is erratic but not completely out.

We accomplished it in this TransitMaster outage scenario by updating concentrate to drop Swiftly bus predictions, but we could extend it to drop from the different modes, or maybe even by route on subway.

Signs-UI is something like this specifically for the countdown signs, but in an ideal world, disabling a station would zero out live predictions in the data feed for all consumers.

## Updates to RTR

I alluded to some updates in RTR. However, I think there are large changes we will need to do either way, and it's somewhat orthogonal to this proposal, so I will leave it off for now.

# Where to go from here?

I think in terms of level of effort we're willing to spend, we have a layered approach:

## Status Quo

Everything _does_ work, nothing is urgently on fire, and "Service Changes" type work can be done by the respective teams individually.

## Concentrate is Keyser Soze

We can improve the internal developer experience by having Concentrate serve as the source of truth, outputting different files for the different internal apps.

Concentrate already _does_ have all the data, so it wouldn't be an enormous lift. We'd have to teach it about non-public fields like operator information, and update it to have different output adapters, and specify which "filters" (pipeline steps) run for each output.

One of the things that excites me about the proposal at a high level is better visibility into our realtime data ecosystem. I don't know that Concentrate is the best platform if we want to continue investing there.

We can improve things certainly, in particular the little devops papercuts that come up with stale feeds or missing predictions, which could be a big help.

But in terms of viewing and slicing of data from the different sources, the internal data model of Concentrate isn't really conducive for that. But we don't really have concrete use cases for these more exploratory ideas anyway. I imagine they would be helpful in debugging and improving and iterating on our realtime data, but that's just a hunch. A lot of the debugging needs to be with _archived_ data, to respond
to past complaints, which means having great visibility into _current_ data is only half the battle.

## Data Visibility B (part of monolith)

If we want to invest a lot in being able to see what all the pieces of our realtime data ecosystem are doing at any moment, we could build a monolithic app. This is closest to the original CentralStation proposal, though it omits the "trip tagging", and potentially "service changes" parts.

(Unfortunately: TODO - a diagram with details on data flows)

This app would have internal state storing the latest feeds from TransitMaster, OCC, Keolis. That data would be serialized for Swiftly. It would have internal state for the output of Swiftly. It would have another store of the data "so far" and serialize that into feeds for concentrate, and have internal state representing what came back from Concentrate. It would combine all this to produce the canonical feeds for Glides, Skate, and the public.

Unfortunately, that's quite a complicated mess of interdependencies so we'd have to be very careful to diagram the data flows, which I sadly haven't had time to do yet, but certainly would if we were even considering this possibility.

Having all the versions of all the data in one place would enable clear debugging visualizations and dashboards.

Performance concerns and latency would have to be evaluated.

Over time, concentrate would be subsumed into this app. Over time CRB and others would be subsumed.

## Data Visibility A (separate tool)

We could also invest in debugging and visualization tooling _without_ actually owning the data. An app could load all the various intermediary app outputs, and let us explore them that way.

paulswartz:
Data Visibility could also be a part of Lamp.

lemald:
Thanks for the big update! Lots of thoughts and questions, but one particular one:

> Trip tagging could be done better with run and block information. Is this generally available?

Do you mean the schedule of the runs and blocks, or the realtime information of what run / block a given vehicle is currently on? In bus land we have both though, the former from the HASTUS schedule and the latter (well, run ID at least) from operators logging in.

lemald:
Re: what sorts of service adjustments Skate users would be entering - good question! You can actually get a sense of what sorts of issues they already log in TransitMaster because those do make it to Skate, on ghosts or buses with the little "!" icon. The bottom line is that they are mostly logging cases where a trip / group of trips won't happen due to a missing operator, a bus taken out of service with mechanical issues, or something like that. We provide what metadata we have to help ops staff, but from a public-facing predictions perspective it all boils down to the same thing: some trips that we thought are happening, aren't happening (unless contra the note in TM a bus does actually log in and run that service).

Paul put together [a doc](https://mbta.sharepoint.com/:w:/r/sites/CTD/_layouts/15/Doc.aspx?sourcedoc=%7BB15913BA-15FD-4632-9871-F192CE75C7F7%7D&file=Swiftly%20Adjustments%20API%20integration%20proposal.docx&action=default&mobileredirect=true) on the Swiftly API a while back - I think it would mostly be explicitly giving data we already have to Swiftly so that they can factor it into their model, rather than us just doing ad-hoc things to the predictions they give back, as well as actually supporting block overloads (which we currently use TransitMaster's predictions for, and those aren't great).

skyqrose:
**What service changes would Glides users input?**

- Trip assignments
- Dropping or adding trips (including adding run-as-directed trips, e.g. for a diversion)
- Changing a trip's destination, or marking it express
- Saying when a trip will depart, or how long it will hold at a station.

Either Central Station or RTR would need to build an API to receive this. I think it'd be easier for Central Station, since it'd be built to take in new and various data sources, but it doesn't make a big difference to the Glides team which project we're sending that to (as long as we don't have to forge OCS messages).

We're hoping to be ready to start publishing data for this in 6mo-1yr.

Edit: We don't care about the way we send data to RTR/Central Station. push or pull or stream, service changes or custom API. Writing the data is a lot easier than consuming it. We can let TRC decide the format that's best for them, and adapt to their needs.

skyqrose:
**Centralization of realtime data**

> @skyqrose - can you provide an example of some issues you've seen with the current [uncentralized] realtime data ecosystem?

Your examples cover it. The two I was thinking of:
- If Swiftly and Busloc disagree on trip assigments, which should we trust? Busloc's are generally lower quality, but Swiftly doesn't understand some of our situations.
- (Hypothetical) If a vehicle is in System A, which updates every 5 seconds, but not System B, which updates every 15, is just because it's a new vehicle and System B hasn't refreshed yet, or is it because A and B disagree about what the vehicle is doing?

These are all solvable problems, but merging is hard, and we should do it exactly once in a centralized place, so that downstream applications don't need knowledge of the upstream parts of the pipeline. Applications should just get data they can trust, so those teams can focus just on their own application logic.

skyqrose:
**Data format**

> It would output different feeds - a glides.json with the all the light rail data including non-revenue vehicles, a skate.json with all the bus vehicles including operator data, and the public data feeds, in the format of GTFS-rt.

I like this, but @digitalcora mentioned an alternative idea in a conversation to me:

Central Station could output a single internal format, that contains all the internal fields it knows about for all modes. e.g. Glides wants an ungeofenced GPS latlng, so Central station would include that field for all modes and requests (though it might be missing/null/default to geofenced GPS for all but Green Line). Then each app could query for data by mode/route. This would reduce (but not eliminate) the Central Station team's burden in supporting new apps. Adding new kinds of data would still require effort, but a new app hitting existing data wouldn't require the Central Station team to add a new-app.json output.

ldnash:
I have been aggressively lurking on this amazing thread. I like the direction this is going and have a few thoughts in response to @losvedir 's most recent update. 

I like the proposed iterative approach with Concentrate. I see the points about the difficulty for introspection, but I also fret about our bad experience with relational databases and performance. As @paulswartz points out, it seems like we could handle this with Lamp, where there's no pressure to process the data as fast as we possibly can.

**ServiceChanges**
Given what's happening on Skate, Glides, and CR maybe this could be an early focus of CentralStation. I (relatedly) think we push Swiftly to accept a ServiceChanges feed rather than making us use their API (they already seem open to this, it's just that few/none of their customers have this capability). It sounds like Concentrate/CentralStation would be the funnel that Skate/Glides/Keolis all feed their ServiceChanges into?

In terms of timing, this is my understanding of where we are with ServiceChanges-y possibilities:

- Now: Bus block waivers from TransitMaster. @lemald pointed out the weird way these are bypassing Swiftly right now. It works and it doesn't have to change right away, but redoing this data flow as ServiceChanges would be a good way to get our hands dirty before we start getting actually new data. 
- Summer: Keolis's updates to vehicle assignments. I won't hazard to guess at a timeline for their longer-term desires. 
- Fall: Skate takes TransitMaster block waivers. In addition to what block waivers are already doing, Skate will also let officials make the more granular adjustments (e.g., to departure times) similar to what @skyqrose mentioned for Glides. 
- Fall/Winter: Glides trainsheets. I suspect it will just be trip adjustments at first, but the other items Sky mentioned will probably come in much the same format if not at the same time. 
- ???: The "last bus of the night problem" would probably make sense to solve with ServiceChanges (e.g., adjust bus departure times based on subway prediction data) but also doesn't have a required timeline.

lemald:
> The "last bus of the night problem" would probably make sense to solve with ServiceChanges (e.g., adjust bus departure times based on subway prediction data) but also doesn't have a required timeline.

@ldnash maybe veering slightly off-track for this PR, but would it make sense to solve this via manually making service adjustments, or would it be better to encode the "w" trips from the bus schedule in GTFS trip-to-trip transfers to indicate that those buses hold for the corresponding last rail trips? Then we can work to improve predictions for the last train trips in RTR, at which point we can point Swiftly to the combination of our static and realtime GTFS and ask them to factor this combined data into their predictions? This also has the advantage of potentially benefiting other data consumers in various ways.

ldnash:
> @ldnash maybe veering slightly off-track for this PR, but would it make sense to solve this via manually making service adjustments

I was actually thinking service adjustments that we create programmatically (rather than manually by officials), which might work similarly to how you describe except we don't need to ask Swiftly for any customizations (which they tend to be resistant to). That said, I think there's probably plenty of ways we could approach this!

lemald:
@ldnash gotcha - I guess my thought was that trip-to-trip transfers aren't quite part of the GTFS base standard but it seems to be one of the [more official extensions](https://developers.google.com/transit/gtfs/reference/gtfs-extensions#TripToTripTransfers), so hopefully that solution wouldn't be regarded as particular "custom," but that may be wishful thinking on my part. 🙂

losvedir:
Unfortunately, I once again find myself without time to wrap up these thoughts as concisely and clearly as possible, but still think I should provide an update on where I feel this work is heading.

The **tl;dr** is this:

* What we've been calling "ServiceChanges" after the proposed GTFS spec, I think we should call "Realtime Events". Something like "holding at a stop" isn't exactly a *service* change, but an update we wish to make to the realtime data.
* The GTFS "ServiceChanges" spec doesn't map cleanly to all our desired use cases, and it's unclear whether we're talking about using it as an interchange format with Swiftly and RTR or publishing it to the public, and it can be confusing if we do some of both.
* I propose Glides, Skate, and Keolis operating on the abstraction level of "events" which we can define based on operational needs like "cancel trip XYZ", "cancel block ABC", "hold at station X for Y mins", which they will emit to Kinesis.
* CentralStation will be the app that consumes these events and updates our data as necessary, whether that's directly, via Concentrate, via Swiftly, or publishing a ServiceChanges feed.
* In the future, we can move more of our realtime system to be message-broker based and have vehicle movement events and the like.

The "single source of truth" aspect of this original RFC is somewhat orthogonal. From mapping out use cases, I think that for CentralStation to effectively handle all the Realtime Events it may need to know about all the vehicle positions and trip updates we're making, so over time this still could be the source of truth, producing internal feeds for the other apps. Or it could feed into concentrate which would become the source of truth. Still TBD since I spent most of the last set of brain cycles on the perhaps more pressing "realtime events" side of things.

# Limitations of "Service Changes"

I have been talking about "service changes". What do I mean by that? I think: capture known deviations from our scheduled (or predicted) service to incorporate into our realtime data. I believe this is an important set of work for CTD in the coming year, and we have a mandate to work on it. However, the GTFS-ServiceChanges spec that I've mentioned, perhaps as more of an implementation detail, I don't think is the best interchange format for that anymore. I will try to call the concept "Realtime Events".

The vision outlined in the CentralStation RFC thus far, and follow-up comments so far was to provide a ServiceChanges feed to Swiftly. However, as I've thought through a lot of the concrete events we want to handle, I foresee a few issues:

* I've muddled whether we're using the GTFS proposal as an interchange format to the predictions providers, or if we're actually publishing them to the public, as intended by the GTFS proposal. If we were to say that a trip was held at a certain stop, we would want *Swiftly* to use that information to update its predictions, but we would *not* want that data to be in a public feed. If Swiftly has already incorporated the information, it might be confusing for the public to consume it as well. On the other hand, something like a change to a CR trip's Flagstop status could make sense in a ServiceChanges StopTimeProperties update to pickup_type and drop_off_type.
* Some of the things we want to do, like "cancel a block" or "cancel a run" aren't natively supported by ServiceChanges.
* I think we might be using ServiceChanges wrong. In general, it seems to target somewhere between realtime data and gtfs_creator changes. GTFS-NewTrips, for example, which "allows adding new trips in real-time, which is useful for replacement shuttles, or for unplanned event-specific shuttles" has a "service level objective" of "ingesting time" on the consumer side of 1-2 hours. If we wanted to publish predictions for a new overload trip, that might be better as a simple ADDED trip using the current TripUpdates spec.
* We've used GTFS-rt as the de facto interchange format for our apps to date, but have had to "enhance" it in various ways. Let's not go down that route again; having a custom internal format with the data we need (and not the data we don't) could make inter-app communication smoother.

My new proposal is that we will orient this aspect of our realtime data (and potentially even vehicle movements in the future) around a message broker architecture. Client apps (Skate, Glides, Keolis) will operate at the abstraction level of "CloudEvent", and we'll have events for all the kinds of modifications they'd like to do: "cancel a trip", "hold a train at a station", etc. They would emit these events to Kinesis streams, and CentralStation will subscribe to the streams and process the events.

# What are the things we want to do, and how would we do them?

## Update trip assignment for a vehicle

The use case here is if the underlying vehicle location feed has an incorrect trip assignment for a vehicle, and someone in the field sees it, they can assign it to a correct trip. Mentioned by: Keolis/Glides.

For Keolis, the trip could be assigned incorrectly in two ways: 1) Clever Devices internal mapping is wrong, perhaps because of an incorrect log-on, or 2) the train is in GPS fallback mode and the Swiftly auto-assigner put it on the wrong trip.

Keolis would observe the incorrect assignment on its internal, stateless "KVL" dashboard which displays data merged from Clever and the MBTA public feed, monitored by customer support staff. If it's a Clever issue, then it would send an internal request to update the assignment. If it's a Swiftly issue, they would like to fire an event that updates the Swiftly assignment.

Swiftly can be updated in one of two ways currently: the AVL feed it consumes, or its assign-vehicles API call. Both have pros and cons. Updating the AVL feed means CentralStation would have to be in the middle of Keolis and Swiftly, and it also means we'd have to worry about *unassigning* the vehicle when it completes the trip. For simplicity, then, I propose CentralStation listens for the "assign trip" event and turns it into a Swiftly API call. One limitation here is that the Swiftly API cannot overrule a vehicle with an explicitly assigned trip in the AVL feed, but that's not an issue here because explicitly incorrect assignments would be fixed internally by Keolis.

For Glides, the assignment is done initially by RTR. With light rail, RTR gets TMOV messages saying what AVI route (the vehicle's EB and WB endpoints) the vehicle is on, and it maintains state about our published GTFS trips via APIv3. RTR tries to infer the best GTFS trip based on the vehicle's pattern, and the time it's at a station. It then assigns either an appropriate "unused" trip to the vehicle, or creates an "ADDED" one. While "block ID" is available, we've never used it.

What exactly is the Glides product flow here? Would users want to assign a vehicle to a different GTFS trip? Would users want to assign the vehicle to a different *route*? If the latter, we'd have to work out the interaction with the vehicle's AVI code. Would users be assigning *future* trips to a vehicle, or generally updating a vehicle's block and/or run?

In this case, CentralStation would handle the "assign trip" event by storing it in its state, and emitting an S3 JSON file that RTR would consume.

This was not a functionality called out by Skate. However, I imagine for incorrect assignments, the desire would be to fix it in TransitMaster?

## Cancel a trip

This is identifying a scheduled trip and indicating that it won't run, ensuring no vehicle is assigned to it, and no predictions are or will be made for it. Mentioned by: keolis/glides/skate

Ultimately we would want our predictions feed to publish a TripUpdate with schedule_relationship of CANCELED as soon as the cancellation is known, and maintain it in the feed until after the scheduled trip has passed.

For Keolis, this would be an event emitted for a given trip. We would use Swiftly's `CANCEL_BLOCK_TRIPS` ServiceAdjustment call. I'm not sure how that manifests in the TripUpdates it publishes, but if we're not happy with it, we could use this event to remove Swiftly's predictions and publish the canceled TripUpdate ourselves.

For Glides, we could either do the same concentrate trick, or we would add this to the S3 JSON file mentioned in trip assignments section above. RTR would not make predictions and would publish the canceled trip per that feed.

Bus already does this; waivers are added to TransitMaster and pulled out by Busloc and turned into StopTimeUpdates, which concentrate uses to suppress the prdictions, and which Skate has to re-assemble into waivers for display. With this proposal, we would instead have Busloc emit a cancel trip event, and CentralStation would receive, store, and turn into StopTimeUpdates. Skate could independently subscribe to the events and maintain its own state with what it needs to know about waivers.

(NB: Is it duplicate work for Skate to maintain waivers in its own DB? The alternative would be to get the current state from CentralStation, but that seems to complicate things a bit and implies an API web server. If we have that, we could even forego all the Kinesis events and have client apps simply POST to the web server.)

## Cancel a block

Called out by Skate, for example to handle a bus with mechanical issues. Swiftly allows you to cancel a whole block with their `CANCEL_BLOCK` API call, but I'm not sure how that ends up in the TripUpdates feed. There's no native support for this in GTFS-rt or GTFS-ServiceChanges, but ultimately it can be just several trips canceled as in "cancel a trip".

This doesn't really apply to Keolis; to Swiftly every CR trip is its own block.

This might be something Glides is interested in. RTR doesn't currently have a concept of "blocks", though it is in the underlying GTFS feed for light rail. I don't know how closely it's followed operationally. If Glides were to emit a "cancel a block" event, CentralStation could use GTFS to enumerate the remaining trips for that block, and cancel each one individually, per the "cancel a trip" process above.

For bus, waivers already support canceling a block. I'm not sure how such an event currently manifests in our predictions. However, we *could* do this at the Swiftly level. If Busloc emitted a "cancel block" event, CentralStation could use Swiftly's API to `CANCEL_BLOCK`.

## Cancel a run

Called out by Skate, for example to handle a missing operator. There's no native support for this by Swiftly or GTFS. Ultimately, though, like with canceling a block, we'd just want to cancel several related trips in our feed.

Keolis didn't bring this up, and I don't know what sort of run data they have that CentralStation could consume for it to calculate the trips belonging to the run.

Glides didn't bring this up, and I don't know if we could make it work. I don't see run information in our HASTUS data in gtfs_creator, but maybe it would be available to us if we asked.

For bus, I'm not sure if this can come through in waivers. Busloc doesn't appear to handle a run_id in its Waiver parsing. Busloc does get the run_id from TransitMaster from operators logging in, but I'm not sure what's involved in connecting that to a set of trips.

(NB: if we wanted to store run information for trips, that would imply storing an "enhanced" version of *static* GTFS.)

## "Pocket" cancel a trip

Akin to a "pocket veto", do we need to worry about explicitly canceling all scheduled trips that simply don't get assignments? RTR does this, but it would make more sense to do it system wide in one place. Concentrate might be a good spot, but we'd need to add a "feed filter" concept perhaps; I think concentrate can modify trips it receives from upstream, but can it add new TripUpdates by itself? Alternatively, if CentralStation is at a point where it knows all the GTFS trips and all the VehiclePositions, it could publish all the cancelled trips.

## ADDED trips (overload/rescue)

This refers to making predictions for trips that aren't in the static schedule, running normal routes just with extra vehicles. This can be done with current GTFS-rt by including trips with a schedule_status of ADDED. In ServiceChanges this has been deprecated and is called DUPLICATED. In either case, it references an existing trip as a base, providing a new start date and time for it. The API does some work to include these ADDED trips into the "static" /trips data, for example giving them route_patterns.

Bus calls this "block overload" and is data we get from TransitMaster, along with some level of predictions, which we currently flow into the public feed. We could instead have Busloc emit a "new trip" event, which CentralStation would read to hit the Swiftly API.

Swiftly supports this through its `DUPLICATE_TRIP` ServiceAdjustment API, which takes a base trip and a new time, and returns an "adjustment_id", which can be used to fetch the new trip_id.

Does TransitMaster generate a trip ID here? If we were to use Swiftly for block overload predictions, I think the flow would be: TransitMaster says there's a block overload for vehicle A with new trip ID B -> CentralStation tells Swiftly to create a duplicate trip, gets back trip ID B2 -> the AVL feed Swiftly consumes has vehicle A updated to be on trip ID B2 (or possibly has "B" cleared so Swiftly auto-assigns B2, though I think we might have auto-assignment disabled for bus?). Does operations use TransitMaster's trip ID? Would it be a problem for the trip ID in our public feed, and presumably Skate, to be different from TransitMaster's internal trip ID?

Glides also brought up ADDED trips, with the extra request of "run-as-directed", for example during a diversion. ADDED/DUPLICATED trips can only reference other trips in the static schedule, so they might not work here (although we could do as RTR does now; generating a new larger trip, e.g. Green-B, and then having an Alert chop off the ends). This would be a place for GTFS-NewTrips, which RTR could produce, and then reference via the TripUpdates. This would mean updates to the API as well, for how it puts the new trips into the static /trips data.

## Held trips

Held trips are a generalization of "train sheets". Train sheets can specify when a given trip will leave its initial terminus, but "holds" can apply at any station down the line. Holds could also be negative, meaning depart early.

Just the train sheets maps neatly into GTFS-rt. The trip can have a StopTimeEvent of either "delay" or "time". And sufficiently future trips can use GTFS-ServiceChanges to specify a StopTimeEvent of "schedule_time". However, for mid-trip holds this doesn't work as well, as operationally the holds are given as a length of time to wait at a station. A "hold for 5 minutes" in the middle of the line doesn't refer to the static schedule, but rather an offset to the predicted schedule, which GTFS-rt doesn't support. Skate or Glides could *calculate* this by consuming the predictions, calculating the offset via the hold length, and then emitting the absolute time or delay relative to the schedule. However, that sounds like the job for CentralStation, and we should expect the operational apps to simply convey what's happening at the operational level of abstraction, via a "hold 5 mins at XYZ" event. One concern with that approach is duplicate events - we'd want to perhaps timestamp the event so we don't accidentally apply "hold 5 mins" twice. Alternatively, we could say holds don't "stack", and that the client apps are responsible for updating "hold 5 mins" to a "hold 10 mins" event.

Swiftly has a `MODIFY_DEPARTURE_TIME` API endpoint which could be used here, though like static data it uses an absolute departure time, so CentralStation would have to calculate that based on current predictions and the given hold time.

RTR could consume a "hold" event in the feed that CentralStation is producing, and use it directly in the prediction making.

## Others I didn't have a chance to flesh out yet

### Local to Express
* GTFS/ServiceChanges
  - StopTimeProperties pickup_type = NO_PICKUP / drop_off_type = NO_DROP_OFF
  - or Standard GTFS-rt with stop time updates being marked Skipped
    - can we mark _just_ the expressed ones skipped for Swiftly to consume? would affect predictions at stops past the express section (train doesn't dwell at stations)
* Swiftly API
  - not directly possible?
  - maybe doing DuplicateTrip with a given known express trip

### Express to Local
* GTFS/ServiceChanges
  - Canceled trip + Added Trip
* Swiftly API
  - not directly possible
  - cancel trip + duplicate trip

### Changes to Flagstop status
* GTFS/ServiceChanges
  - StopTimeProperties
  - Unlikely to affect predictions? Would this be a truly separate piece of data in ServiceChanges.pb then?
* Swiftly API
  - not possible? I guess cancel trip + duplicate trip if another trip has same flag stop pattern

lemald:
> "Pocket" cancel a trip
>
> Akin to a "pocket veto", do we need to worry about explicitly canceling all scheduled trips that simply don't get assignments? RTR does this, but it would make more sense to do it system wide in one place.

One potentially complicating factor here: across different modes, the fact that we don't have active tracking for a vehicle can give us differing levels of certainty about the fact that a trip isn't operating depending on how reliable the tracking methods in question are. Still, maybe it would make sense to have Concentrate or CentralStation be the application implementing this and we could just enable / disable it on a per-incoming-feed basis.

lemald:
Also quick note on "block waivers," despite the name they don't necessarily apply to a whole block - there's some flexibility to select a start and end to the duration of the waiver, and BusLoc uses this information to publish StopTimeUpdates only for those stop times falling within that range.

boringcactus:
> The GTFS "ServiceChanges" spec doesn't map cleanly to all our desired use cases, and it's unclear whether we're talking about using it as an interchange format with Swiftly and RTR or publishing it to the public, and it can be confusing if we do some of both.

I have been talking about GTFS-ServiceChanges solely as an interchange format. For Glides in particular, the only event which doesn't fit within the existing GTFS-realtime format is when a trip is extended to stops it wouldn't normally visit; everything else fits nicely into GTFS-realtime as it already exists.

> What exactly is the Glides product flow here [for trip assignment]? Would users want to assign a vehicle to a different GTFS trip? Would users want to assign the vehicle to a different route? If the latter, we'd have to work out the interaction with the vehicle's AVI code. Would users be assigning future trips to a vehicle, or generally updating a vehicle's block and/or run?

Trip assignments are always entered into Glides trainsheets before the trip has started. Trainsheets are the source of truth there, so RTR will no longer need to guess trip assignments once Glides has trainsheet data.

> A "hold for 5 minutes" in the middle of the line doesn't refer to the static schedule, but rather an offset to the predicted schedule, which GTFS-rt doesn't support.

I don't know if this is the only way to think of this. From my perspective, static GTFS defines the theoretical expectations for arrival and departure times, and static GTFS plus Glides trainsheet data defines the actual expectations for arrival and departure times, so "hold for 5 minutes" means the actual expectation is five minutes later than the theoretical expectation, which can be defined using the GTFS-realtime delay field.

lemald:
> NB: Is it duplicate work for Skate to maintain waivers in its own DB?

I'm fine with maintaining the existing setup where Skate saves waiver information to the database as "notifications." Currently those are the main (but not only) source of notifications, and we might add more in the future, and if Skate is turning different kinds of external input into an internal concept of user notifications I think it makes sense for us to own whatever persistence of those is necessary.

digitalcora:
To the technical point of using Kinesis: It seems like we expect CentralStation would consume all such events, and that it would be the only consumer of such events (if others were consuming the raw events directly, then CS would not be doing its job as a single-source-of-truth). In that case, why _not_...

> forego all the Kinesis events and have client apps simply POST to the web server

...? i.o.w. what capabilities are unlocked in exchange for the increased complexity of using an external event bus? Not advocating for one approach over the other, just interested in the contributing factors.

> (NB: if we wanted to store run information for trips, that would imply storing an "enhanced" version of _static_ GTFS.)

I have a truly marvelous proposal for a combined CentralStation / API v4 / gtfs_creator3 along these lines, which this margin is too narrow to contain 😉

lemald:
I think @losvedir's big comment going through specific thorny cases is really useful. It also makes me realize that I should probably make an attempt at listing out my high-level requirements from the Skate / Bus Tech end:

1. The raw information about the service adjustment as entered by the user should be easily recovered from the interchange format, not just the aspects of it that have an immediate effect on rider-facing predictions. This is in order to facilitate integration with other operations tools or sources of operations information (maybe Lamp would want to record some of this information?).
2. The format should be relatively free-standing on not rely on an additional implicit contract between the producer and consumer as to how to interpret certain pieces of information. The unfortunate thing is that many parts of GTFS-rt are a little under-specified as-is even for their intended use in passenger-facing communications, so we already have some workflows that rely on somewhat underhanded data pipelines with different components combining different GTFS-rt feeds in unorthodox ways.
3. I agree this is mostly orthogonal to whether or not TRC does CentralStation.
4. Regardless of software architecture around CentralStation, we should be clear about where the source of truth for different kinds of data lives. Applications (CentralStation or not) within TRC's purview are probably going to continue being the source of truth for real-time rider information, but what about internal ops data that doesn't show up in said rider information for various reasons? Do Skate and Glides just store this in their own respective databases? Does it go into some shared place like Lamp? Does it live with TRC team somehow after all?

arkadyan:
@lemald Is there a need in bus land for changes to be written back upstream into the TransitMaster database? My assumption is that we would not expect that tool to be updated to consume any of our data sources, but I could be off base there.

lemald:
> @lemald Is there a need in bus land for changes to be written back upstream into the TransitMaster database? My assumption is that we would not expect that tool to be updated to consume any of our data sources, but I could be off base there.

That's actually a "maybe" at this point. If we do end up putting data back into TM, though, it will be via an HTTP API.

skyqrose:
**Kinesis**: +1 for wondering why it should go through Kinesis if we have only one consumer.

**Runs/Blocks**: Glides should not need to cancel a run or a block all at once. Blocks aren't used operationally; inspectors figure out what train to use trip by trip. Runs are used (and Glides exports them from HASTUS), but if an operator is out, inspectors will try to fill their trips one by one with swaps or with the cover list, so wouldn't want to cancel the whole run at once. And even if they do, the Glides app can do that by sending a bunch of separate trip cancellations without Central Station needing to support runs.

**Added trips**: Yes, Glides wants to be able to add trips that don't follow an existing trip.

PR #13, opened 2022-05-16
consistent resource naming RFC

paulswartz:
Asana task: https://app.asana.com/0/1113179098808463/1202191100361265/f


rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+As CTD has grown, multiple teams have independently named their projects and the associated resources. As these names have been generated independently, they are not always consistent about how they use punctuation to separate multiple words in a resource name, leading to confusion.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Within CTD, we use one of the following patterns for naming resources and :
```
losvedir: Sentence seems to end prematurely.

rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,94 @@
...
+	- https://peps.python.org/pep-0008/
+- React components: PascalCase
+- TypeScript modules: camelCase
+- Shell scripts: kebab-case with no trailing `.sh`
+- Domain names: kebab-case
+	- https://www.ssl.com/faqs/underscores-not-allowed-in-domain-names/
+- S3 buckets: kebab-case
+	- https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+- Terraform resources: kebab-case
+- Terraform modules: kebab-case
```
losvedir: I'm not sure if it can be otherwise, but terraform _variables_ generally seem to be snake case.

rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,94 @@
...
+- IAM roles: kebab-case
+- IAM groups: kebab-case
+- IAM users: kebab-case  (but prefer IAM roles whenever possible)
+- EC2 security groups: kebab-case
+- Cognito user pools: kebab-case
+- Cognito app clients: kebab-case
+- Glue databases/tables: snake_case
+	- https://docs.aws.amazon.com/athena/latest/ug/glue-best-practices.html#schema-names
+
+Abbreviations should be treated as a single word and have their case ignored: gtfs_documentation rather than GTFSDocumentation.
```
losvedir: Can you clarify what "ignored" means? This was an issue with "Signs UI" and specifically how to name the modules and files. We standardized on `SignsUi`, but I think the Elixir convention is that it should have been `SignsUI`.

skyqrose:
Should this RFC mention whether or not we plan to change existing non-compliant names to match the new format? (Some things will be easy to change, others less so)

rfcs/0000-consistent-resource-names.md
paulswartz: added.

rfcs/0000-consistent-resource-names.md
paulswartz: I meant this, but didn't realize that Elixir prefers the alternate pattern. noted as an exception.

paulswartz:
@skyqrose [it does](https://github.com/mbta/technology-docs/pull/13/files#diff-5af1164513c73a814d48f84359c9563382040fb633a34d72be52d5880d24de4cR62), and we do not:

> This RFC does not intend to force existing applications to rename existing files: only to provide standardization as CTD continues to grow and create new projects.

rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,97 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Within CTD, we use one of the following patterns for naming projects and resources:
+- kebab-case: all lowercase, separated by hyphens
+- snake_case: all lowercase, separated by underscores
+- camelCase: first word is lower case, other words are Title case with no separation
+- PascalCase: all words are Title case with no separation
+
+We use the following patterns for resources:
```
grejdi-mbta: Minor: I would love to move away from `-` for everything. I always find myself having to use a mouse pointer to highlight names with `kebab-case`. With any other case, it's a quick double click, or keyboard shortcuts. Not sure if this is the kind of feedback you're looking for. 😄 

Question:  Does it make sense to indicate what is the current practice and what is recommended?

rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,97 @@
...
+- TypeScript modules: camelCase
+- Shell scripts: kebab-case with no trailing `.sh`
+- Domain names: kebab-case
+	- https://www.ssl.com/faqs/underscores-not-allowed-in-domain-names/
+- S3 buckets: kebab-case
+	- https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+- Terraform resources: kebab-case
+- Terraform modules: kebab-case
+- Terraform variables: snake_case
+- ECS clusters: kebab-case
```
grejdi-mbta: Is it necessary to indicate all of these? Feels very detailed. The only exception seems to be Glue. Maybe it could tucked under `- Terraform resources: kebab-case` above as an exception.

rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,97 @@
...
+[prior-art]: #prior-art
+
+Existing CTD repositories were considered for these recommendations, as were our existing infrastructure. In some cases, language-specific style guides / best practices were also considered.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Are there any other types of resources where we should be standardizing on names?
+- Should we be standardizing even further (Credo configurations, Eslint packages, &c)?
+- Terraform resources use camel_case: should we be using that ourselves?
```
grejdi-mbta: 👍

rfcs/0000-consistent-resource-names.md
```diff
@@ -0,0 +1,97 @@
...
+- React components: PascalCase
+- TypeScript modules: camelCase
+- Shell scripts: kebab-case with no trailing `.sh`
+- Domain names: kebab-case
+	- https://www.ssl.com/faqs/underscores-not-allowed-in-domain-names/
+- S3 buckets: kebab-case
+	- https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+- Terraform resources: kebab-case
+- Terraform modules: kebab-case
+- Terraform variables: snake_case
```
paulswartz: @ianwestcott these (in particular modules) are based on our current practice. However. the larger Terraform ecosystem appears to be using snake_case. Do you have thoughts here?

ianwestcott: It does look like [Terraform's official recommendation is to use `_` (underscore) for all named things](https://www.terraform-best-practices.com/naming), which I was unaware of when I started writing out/importing our configuration. So I'm okay with adopting the official convention, even if it means a lot of what exists now is non-compliant.

rfcs/0000-consistent-resource-names.md
paulswartz: Cleaned up all the Terraform items.

rfcs/0000-consistent-resource-names.md
paulswartz: What is "current practice" will change as the recommendations in this RFC are adopted, so leaving that out allows the document to avoid being out-of-date in the future.

paulswartz:
Per the [README](https://www.terraform-best-practices.com/naming) I'm calling for a final comment period with a disposition to accept. If there are no blocking concerns by 2022-07-07 12:00 EDT, I will consider this accepted and merge.

paulswartz:
RFC is accepted!

PR #15, opened 2022-09-15
rfc: application environment naming

lemald:
[Rendered](https://github.com/mbta/technology-docs/blob/lem-environment-naming/rfcs/0015-consistent-environment-naming.md)


skyqrose:
Yes! I like this, and often call the staging environments "staging" in conversation to disambiguate them from local "dev" environments, (for example, `MIX_ENV=dev` / `config/dev.exs`).

Perhaps we could also encourage describing local environments as `local` rather than `dev`, though that might contradict the MIX_ENV conventions. "dev" is such an overloaded term and it'd be nice to not try to have anything precisely named "dev".

skyqrose:
Switching does seem like a lot of work. I'd at least propose we call them "staging" in conversation even while infrastructure and domain names are still named "dev", and updating the recommendations in the new app checklist.

I don't think mandatory switches are necessary, since we could end up scraping the dregs of high effort low reward apps, but we should at least recommend switching most apps.

Also, the infrastructure team is pretty busy and this would probably fall largely on them to do, so perhaps actively switching apps should be low priority compared to their other work.

lemald:
> Yes! I like this, and often call the staging environments "staging" in conversation to disambiguate them from local "dev" environments, (for example, `MIX_ENV=dev` / `config/dev.exs`).
> 
> Perhaps we could also encourage describing local environments as `local` rather than `dev`, though that might contradict the MIX_ENV conventions. "dev" is such an overloaded term and it'd be nice to not try to have anything precisely named "dev".

Yes, good point, I forgot about "dev" in the Elixir application configuration sense but that is yet another overloaded use of the word, and as you say it seems to be pretty solidified in the MIX_ENV conventions. I don't see any obvious improvements we could make to naming to fix this part but I'll think about it.

skyqrose:
It might be nice if this document had a short guide for how to change an existing application. There might be a lot of places to change, and I would want to update my application but don't quite know where to start.

lemald:
> It might be nice if this document had a short guide for how to change an existing application. There might be a lot of places to change, and I would want to update my application but don't quite know where to start.

That's a good point! One thought I had is that most dev / staging environments don't have a ton of important persistent data associated with them, even if they have a database, so one option that's easy to implement in that case is simply copying over the old configuration in Terraform, renaming it, creating the new environment, and then deleting the old one. Actually renaming the old environment while keeping the old resources in place I'm less sure about - @ianwestcott would you be able to share some thoughts on how that might go?

ianwestcott:
Given that Terraform makes it fairly easy to stand up a new environment with an arbitrary environment name, my official recommendation would be for teams to create new environments that are in line with this naming convention and then tear down their non-compliant environments (at their leisure, of course).

I agree that there are minimal stateful resources to be concerned about, but if there's a database that needs to be carried over from one environment to another, we can handle that via the documented snapshot/restore process: https://github.com/mbta/wiki/blob/master/devops/rds-database-maintenance.md#snapshot--restore That said, ideally every application should have some scaffolding in place to initialize a new database, which would also be useful for local development as well as disaster recovery.

lemald:
Thanks @ianwestcott! I'll add that to the RFC as the recommended approach, along with a link to that wiki page for teams that do want to hang on to some database contents.

rfcs/0015-consistent-environment-naming.md
```diff
@@ -0,0 +1,84 @@
...
+
+- `prod` environments should only ever get their data from other `prod` environments.
+- `staging` environments should only ever get their data from other `staging` environments or `prod` environments, depending on which is most appropriate for monitoring application stability.
+- `test` environments should generally default to getting their data from `staging` environments. In cases where there is some end-to-end test process across multiple applications, it may make sense to have one `test` environment get its data from another application's `test` environment.
+- `dev-[color]` environments should generally default to getting their data from `staging` environments. When multiple teams are working on interrelated features (for instance, TRC is introducing a data source and dotcom is making updates to display that data), it may make sense to point one `dev-[color]` environment to another application's `dev-[color]` environment.
```
beatrixada: question: did you mean to suggest that teams working together should use the same color for environments hosting shared and dependent work? e.g. that TRC and Dotcom should both use their `dev-green` environment for testing to make it easier to remember data flows.

rfcs/0015-consistent-environment-naming.md
```diff
@@ -0,0 +1,84 @@
...
+
+Non-production environments for our applications are generally named with some variation on `dev`. This doesn't capture the distinctions between different non-production uses of application environments.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Environments are named in `kebab-case`, following the convention `[name of application]-[modifier]`. Available modifiers are as follows, with a brief description of their use case:
+
+- `prod`: the production environment.
+- `staging`: used for final integration testing and monitoring for errors before deploying changes to production. Often, merging a PR into the `main` / `master` branch will be set up to automatically initiate a deploy to the application's `staging` environment. There should be one stage environment per application.
```
beatrixada: question: TRC has internal dashboard-style apps like `prediction_analyzer` and `realtime-dashboard` that currently use the `dev` environment to indicate that they are not serving production traffic. `dev` is the only environment for these apps. Does `staging` still make sense as a label?

rfcs/0015-consistent-environment-naming.md
```diff
@@ -0,0 +1,84 @@
+- Feature Name: consistent-environment-naming
+- Start Date: 2022-09-14
+- RFC PR: [mbta/technology-docs#0015](https://github.com/mbta/technology-docs/pull/15)
+- Asana task: [Dev environment naming / usage conventions](https://app.asana.com/0/1200506724882024/1202675232761300/f)
+- Status: Proposed
+
+# Summary
+[summary]: #summary
+
+Non-production environments of CTD applications have clear, descriptive, and consistent naming.
```
beatrixada: praise: thanks for writing this RFC! Hopefully it can alleviate some of the struggle in remembering what flows to what.

rfcs/0015-consistent-environment-naming.md
```diff
@@ -0,0 +1,84 @@
...
+- `test`: used as the target environment for some sort of recurring automated testing, such as end-to-end web application testing or load testing. Accepts an optional modifier specifying what testing tool the environment is for (for instance, `test-cypress` or `test-locust`). This is presently a hypothetical use case, albeit one that has been discussed from time to time.
+- `dev-[color]`: for experimenting with and evaluating features under active development, generally in their own branch. Potential uses include: allowing a developer to experiment with code that relates to functionality that's hard to replicate locally, providing a space for other team members to evaluate and sign off on a feature before it is merged, and collecting application-specific metrics on a particular change before deciding whether to merge it. `[color]` can be the color of any MBTA rapid transit line (`green`, `red`, `orange`, `blue`, or `silver`).
```
beatrixada: question: if in the future we were to have fully ephemeral environments (i.e. the ability for a deployment to spin up a new environment itself for testing a feature) would this naming scheme still work? I worry that we might run out of transit lines and I don't want to artificially limit the number of app environments we can have. Additionally, we're currently discussing manually spinning up a new `prediction_analyzer` environment just for running a specific experiment against Swiftly data. This environment would only exist for the duration of the experiment, as a way of being able to maintain full `prediction_analyzer` capabilities for current service and the experiment concurrently. I would prefer to call this environment `swiftly-experiment` or something like that, would you advise giving it a line color like `dev-green` instead?

rfcs/0015-consistent-environment-naming.md
lemald: Good question - I hadn't specifically had that in mind, though I could imagine cases where it's hard to coordinate this (Team A has already been using dev-green for this project and realizes they need to do some integration testing with Team B, but Team B is already using dev-blue). In any event, I could add a note that it's preferred if possible.

beatrixada: Yeah I'm not sure it's a good idea, I just wanted to clarify. Seems like it could be helpful in some cases and not others.

rfcs/0015-consistent-environment-naming.md
lemald: That's a really good point. I wonder if there's another suffix we could use. Maybe `internal` or something, to indicate that it's not staging but it's still not quite "real" prod? @ianwestcott do you have any opinions on this? Are there any reasons that it would be inappropriate to just put `internal` right there in the domain name?

rfcs/0015-consistent-environment-naming.md
lemald: Thanks! 😀

rfcs/0015-consistent-environment-naming.md
lemald: Also very good questions. This is definitely written under the assumption that dev environments are something that you spin up and that then stick around for a long time, and you reuse them for a variety of purposes as time goes on. I think that describes our current paradigm of operating pretty well, but that doesn't necessarily need to be the case in the future. Maybe I can add a clause along the lines of "...unless it's strictly for testing a specific piece of functionality, in which case feel free to give it a name that indicates what the feature in question is."

skyqrose: One of the benefits of Terraform was to make it easy to spin up single-use environments like this, and I expect it will happen more than once in the future. I think it should still have a prefix, so `dev-swiftly-experiment` or `tmp-swiftly-experiment` or something.

lemald: I pushed a change reflecting that names specifying the feature being tested are allowed. I think they should still be `dev-` rather than `tmp-` or something else like that because `dev-` reflects the functional use of the environment. Separately I'd argue that our Terraform setup isn't quite there yet in terms of making it easy to set up a temporary environment for a feature that's under development for less than a few weeks or so, but that's definitely something we should accommodate in the future.

rfcs/0015-consistent-environment-naming.md
```diff
@@ -0,0 +1,84 @@
...
+
+The main piece of prior art that this RFC draws from is the use of "staging" as terminology to describe the use case for most of our current "dev" environments. This appears to be fairly common in the industry; it was in use at my previous company, and other CTD engineers have understood what I mean by it when discussing this issue.
+
+CTD has already been using `dev-green` as an environment name, as well as `dev-blue` in a handful of cases. This naming dates from when CTD was experimenting with [blue-green deployments](https://martinfowler.com/bliki/BlueGreenDeployment.html), a practice that was not ultimately adopted. Since these are already colors of MBTA rapid transit lines, it can be reworked to provide the basis for an expanded naming scheme that's also thematic. Many teams will probably only need one or two `dev` environments, but larger teams that work on a lot of features simultaneously may benefit from more, and this proposal provides convenient names for them.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Will there be any impact on non-production environments of the various vendors that interact with our data pipiline, like Swiftly?
```
beatrixada: nitpick(typo):
```suggestion
- Will there be any impact on non-production environments of the various vendors that interact with our data pipeline, like Swiftly?
```

lemald:
Going ahead and merging after final discussion with other tech leads. The last remaining area of substantial discussion was around @bklebe's comment about applications without a conventional "prod" environment. The conclusion there was that this is more of a question of how to classify these environments, although in the future we could still adopt an additional naming scheme for these "not-quite-prod-in-the-normal-sense" environments if we want to.

PR #16, opened 2022-11-16
rfc: Terraform Workflow Improvements

ianwestcott:
Asana Task: https://app.asana.com/0/1113179098808463/1202075979488110/f

Rendered version: https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md

### DIFF
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -230,7 +230,7 @@ Once a new root module is set up for each engineering team, we'll need to:
 - Migrate each team's development and staging resources out of the `dev` module and into their new team-based module
 - Update references to dependent resources that remain in other modules to use data sources or [remote state](https://developer.hashicorp.com/terraform/language/state/remote-state-data)
 
-Note that production resources will remain in a single `prod` root module in the short term, but will likely eventually also be moved to team-based modules a later date. This will allow us to vet the changes proposed in this document, while also making sure we're adhering to the stricter security requirements for production that are discussed in the [Scalr workflow for production][scalr-workflow-production] section.
+Note that production resources will remain in a single `prod` root module in the short term, but will likely eventually also be moved to team-based modules at a later date. This stepped approach will allow us to vet the changes proposed in this document, while also making sure we're adhering to the stricter security requirements for production that are discussed in the [Scalr workflow for production][scalr-workflow-production] section.
 
 ### Root Module Naming Convention
 [root-module-naming-convention]: #root-module-naming-convention
@@ -271,7 +271,7 @@ Root modules associated with the "CTD Production" AWS account that should be ren
 - `onprem_prod` → multiple modules:
   - move SSM resources to `aws-ctd-main-base`
   - move applications to `aws-ctd-main-prod` for now
-- `prod` → `aws-ctd-main-prod` (to remain as a single module for now)
+- `prod` → `aws-ctd-main-prod` (to remain as a single module in the short term)
 - `restricted` → `aws-ctd-main-restricted`
 
 Root modules to remain as-is:
@@ -462,7 +462,7 @@ Each root module will have a GitHub-integrated Scalr workspace. Scalr will autom
 
 Production will mirror the development/staging context, with a few notable differences:
 
-1. For the time being, all resources will remain in the central `prod` module, which will be [renamed][root-module-naming-convention] to `aws-ctd-main-prod` and have its own GitHub-integrated Scalr workspace.
+1. In the short term, all resources will remain in the central `prod` module, which will be [renamed][root-module-naming-convention] to `aws-ctd-main-prod` and have its own GitHub-integrated Scalr workspace. Once the workflow has been vetted and security requirements are well understood, we can begin to move production resources into team-based modules.
 2. All calls to un-versioned modules will pinned to particular git hashes instead of relative paths, in order to protect the production context from changes that are made to un-versioned modules.
 
 As before, Scalr will automatically run `terraform plan` when a pull request is opened, and attempt to automatically run `terraform apply` on merge.
@@ -824,7 +824,6 @@ in this or subsequent RFCs. The section merely provides additional information.
 
 Ideas that are out of scope for the work described in this RFC, but are potentially worth implementing later:
 
-- Splitting production resources out of the `prod` module and into team-based root modules
 - Handling the [module testing workflow][reference-module-testing] as a GitHub Actions workflow instead of manually
 - Publishing more of our modules in [Terraform's public registry][terraform-module-publishing]
 - Investigating [Terragrunt](https://terragrunt.gruntwork.io/) as a way to standardize backend and provider configuration and reduce redundant configuration
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -237,7 +237,7 @@ Terraform's official [naming convention for published modules][terraform-module-
 
 - AWS account nickname (e.g. `mbta-ctd-test`, `mbta-ctd-main`, or preferably a more succinct version that eliminates the `mbta-` prefix)
 - Team name (prefixed with `team-`)
-- Environment context (e.g. `dev`, `staging`, `prod`) - this overlaps with AWS account nickname, so we may want to discourage the use of both at the same time
+- Environment context (e.g. `dev`, `staging`, `prod`)
 - Component descriptor of the type/category of resources the module is intended to contain, if it's one of multiple modules for the particular provider/account/context (e.g. `base`, `bootstrap`, `restricted`)
 - Nothing (in cases where one module encompasses all of a provider's configuration)
 
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -266,8 +266,8 @@ Root modules associated with the "CTD Production" AWS account that should be ren
   - move applications to team-based modules
 - `onprem_prod` → multiple modules:
   - move SSM resources to `aws-ctd-main-base`
-  - move applications to `aws-ctd-main-apps` for now
-- `prod` → `aws-ctd-main-apps` (to remain as a single module for now)
+  - move applications to `aws-ctd-main-prod` for now
+- `prod` → `aws-ctd-main-prod` (to remain as a single module for now)
 - `restricted` → `aws-ctd-main-restricted`
 
 Root modules to remain as-is:
@@ -452,7 +452,7 @@ Each root module will have a GitHub-integrated Scalr workspace. Scalr will autom
 
 Production will mirror the development/staging context, with a few notable differences:
 
-1. For the time being, all resources will remain in the central `prod` module, which will be [renamed][root-module-naming-convention] to `aws-ctd-main-apps` and have its own GitHub-integrated Scalr workspace.
+1. For the time being, all resources will remain in the central `prod` module, which will be [renamed][root-module-naming-convention] to `aws-ctd-main-prod` and have its own GitHub-integrated Scalr workspace.
 2. All calls to un-versioned modules will pinned to particular git hashes instead of relative paths, in order to protect the production context from changes that are made to un-versioned modules.
 
 As before, Scalr will automatically run `terraform plan` when a pull request is opened, and attempt to automatically run `terraform apply` on merge.
@@ -460,14 +460,14 @@ As before, Scalr will automatically run `terraform plan` when a pull request is 
 It is a security requirement to have an extra layer of protection on production systems, that any changes to those system require explicit approval. CTD typically handles this requirement during the code review stage. For Terraform changes, we can use the same approach, provided that sufficient protections are in place to guarantee that changes can't circumvent the code review process. Those protections will include:
 
 - [Branch protection rules](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) for the default branch of the devops repo
-- [A `CODEOWNERS` file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) listing the required reviewers for each CTD application's Terraform configuration file within the `aws-ctd-main-apps` (née `prod`) root module
+- [A `CODEOWNERS` file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) listing the required reviewers for each CTD application's Terraform configuration file within the `aws-ctd-main-prod` (née `prod`) root module
 
 The code owners for each application can include engineering team leads, team members, and/or the Infrastructure team.
 
 ##### Continuous Integration &amp; Testing
 
 - Module changes are first [tested in isolation][scalr-workflow-module-testing] and [applied in the development/staging context][scalr-workflow-development-staging]
-- Engineer updates module version or ref in `aws-ctd-main-apps` root module
+- Engineer updates module version or ref in `aws-ctd-main-prod` root module
 - Engineer pushes changes to GitHub branch
 - GitHub Actions workflow runs:
   - `terraform validate`
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -324,7 +324,7 @@ As noted in the [guide-level explanation][guide-module-testing], testing will ha
 
 The test AWS account will have some base resources predefined, such as VPCs and subnets, via the `aws-ctd-test-base` Terraform root module. This module will also define an output a `context` var that test modules can use to reference these shared resources.
 
-In some cases it may also be necessary for test infrastructure to make use of real development/staging infrastructure in the `mbta-ctd-main` AWS account; this can be handled via cross-account IAM roles if necessary. However, the test AWS account should not be considered a safe place for sensitive data, and we'll need to make sure that no PI/PII can leak out of the main AWS account into the test account. For this reason, production resources will likely not be allowed to be shared with resources in the test AWS account, and any data shared between the accounts will need to be scrubbed of PI/PII in transit.
+In some cases there may be a desire for test infrastructure to make use of real development/staging infrastructure in the `mbta-ctd-main` AWS account. In theory this could be handled via cross-account IAM roles. However, the test AWS account should not be considered a safe place for sensitive data, and great care would need to be taken to make sure that no PI/PII can leak out of the main AWS account into the test account. Given this caveat, production resources will generally not be allowed to be shared with resources in the test AWS account, and any data shared between the accounts will need to be copied over and scrubbed of PI/PII in transit.
 
 ## 3. Standardizing on Infrastructure Component Modules
 [reference-standardization]: #3-standardizing-on-infrastructure-component-modules
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -235,42 +235,48 @@ We'll also need to implement a new naming convention for root modules to keep th
 
 Terraform's official [naming convention for published modules][terraform-module-publishing] is `terraform-<PROVIDER>-<NAME>`, where `<NAME>` can contain additional hyphens. Since our own root modules won't be published to Terraform's public registry, we can omit the `terraform-` prefix and use `<PROVIDER>-<NAME>` as a base convention, where `<NAME>` can contain any of the following:
 
-- AWS account nickname (e.g. `mbta-ctd-test`, `mbta-ctd-prod`, or preferably a more succinct version that eliminates the `mbta-` prefix)
+- AWS account nickname (e.g. `mbta-ctd-test`, `mbta-ctd-main`, or preferably a more succinct version that eliminates the `mbta-` prefix)
 - Team name (prefixed with `team-`)
 - Environment context (e.g. `dev`, `staging`, `prod`) - this overlaps with AWS account nickname, so we may want to discourage the use of both at the same time
 - Component descriptor of the type/category of resources the module is intended to contain, if it's one of multiple modules for the particular provider/account/context (e.g. `base`, `bootstrap`, `restricted`)
 - Nothing (in cases where one module encompasses all of a provider's configuration)
 
-Root modules associated with the "CTD Production" AWS account that should be renamed: _(note that our production AWS account is currently named "mbtacomsupport", but we are hoping to rename it to "mbta-ctd-prod" soon)_
+A side-note regarding the AWS account nicknames: Our main AWS account is currently named "mbtacomsupport", but we are hoping to [rename it](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-alias.html) soon. We toyed with the name "mbta-ctd-prod", but since the account also contains development/staging resources&mdash;which can't reasonably be provisioned into their own separate "development/staging" AWS account&mdash;we've opted to recommend the name "mbta-ctd-main". That name is open to debate, however.
 
-- `bootstrap` → `aws-ctd-prod-bootstrap`
-- `global` → `aws-ctd-prod-base`
+#### Proposed Module Names
+
+With all the above in mind, the following are the proposed names for all existing and future root modules. (Note that the naming convention for child modules is different, and is covered in the [module versioning][guide-module-versioning] section.)
+
+Root modules associated with the "CTD Production" AWS account that should be renamed:
+
+- `bootstrap` → `aws-ctd-main-bootstrap`
+- `global` → `aws-ctd-main-base`
 - `dev` → multiple modules:
   - move application development/staging resources to team-based modules:
-    - `aws-ctd-prod-team-dataplatform`
-    - `aws-ctd-prod-team-glides`
-    - `aws-ctd-prod-team-lamp`
-    - `aws-ctd-prod-team-pass-programs`
-    - `aws-ctd-prod-team-screens`
-    - `aws-ctd-prod-team-trc`
-    - `aws-ctd-prod-team-website`
-  - move base dev resources to `aws-ctd-prod-base`
+    - `aws-ctd-main-team-dataplatform`
+    - `aws-ctd-main-team-glides`
+    - `aws-ctd-main-team-lamp`
+    - `aws-ctd-main-team-pass-programs`
+    - `aws-ctd-main-team-screens`
+    - `aws-ctd-main-team-trc`
+    - `aws-ctd-main-team-website`
+  - move base dev resources to `aws-ctd-main-base`
 - `onprem_dev` → `aws-onprem-dev`
 - `onprem_prod` → `aws-onprem-prod`
-- `prod` → `aws-ctd-prod-apps` (to remain as a single module for now)
-- `restricted` → `aws-ctd-prod-restricted`
+- `prod` → `aws-ctd-main-apps` (to remain as a single module for now)
+- `restricted` → `aws-ctd-main-restricted`
 
-Modules to remain as-is:
+Root modules to remain as-is:
 
 - `scalr`
 
-New modules for "mbta-ctd-test" AWS account:
+New root modules for "mbta-ctd-test" AWS account:
 
 - `aws-ctd-test-base`
 - `aws-ctd-test-bootstrap`
 - `aws-ctd-test-restricted`
 
-Possible future modules:
+Possible future root modules:
 
 - `github`
 - `sentry`
@@ -442,7 +448,7 @@ Each root module will have a GitHub-integrated Scalr workspace. Scalr will autom
 
 Production will mirror the development/staging context, with a few notable differences:
 
-1. For the time being, all resources will remain in the central `prod` module, which will be [renamed][root-module-naming-convention] to `aws-ctd-prod-apps` and have its own GitHub-integrated Scalr workspace.
+1. For the time being, all resources will remain in the central `prod` module, which will be [renamed][root-module-naming-convention] to `aws-ctd-main-apps` and have its own GitHub-integrated Scalr workspace.
 2. All calls to un-versioned modules will pinned to particular git hashes instead of relative paths, in order to protect the production context from changes that are made to un-versioned modules.
 
 As before, Scalr will automatically run `terraform plan` when a pull request is opened, and attempt to automatically run `terraform apply` on merge.
@@ -450,14 +456,14 @@ As before, Scalr will automatically run `terraform plan` when a pull request is 
 It is a security requirement to have an extra layer of protection on production systems, that any changes to those system require explicit approval. CTD typically handles this requirement during the code review stage. For Terraform changes, we can use the same approach, provided that sufficient protections are in place to guarantee that changes can't circumvent the code review process. Those protections will include:
 
 - [Branch protection rules](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches) for the default branch of the devops repo
-- [A `CODEOWNERS` file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) listing the required reviewers for each CTD application's Terraform configuration file within the `aws-ctd-prod-apps` (née `prod`) root module
+- [A `CODEOWNERS` file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) listing the required reviewers for each CTD application's Terraform configuration file within the `aws-ctd-main-apps` (née `prod`) root module
 
 The code owners for each application can include engineering team leads, team members, and/or the Infrastructure team.
 
 ##### Continuous Integration &amp; Testing
 
 - Module changes are first [tested in isolation][scalr-workflow-module-testing] and [applied in the development/staging context][scalr-workflow-development-staging]
-- Engineer updates module version or ref in `aws-ctd-prod-apps` root module
+- Engineer updates module version or ref in `aws-ctd-main-apps` root module
 - Engineer pushes changes to GitHub branch
 - GitHub Actions workflow runs:
   - `terraform validate`
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -581,9 +581,9 @@ This list may not be comprehensive; feel free to suggest additional alternatives
 ## Alternatives to Splitting Root Modules
 [alternatives-module-splitting]: #alternatives-to-splitting-root-modules
 
-We feel the [proposed approach][guide-module-splitting] adds a lot of flexibility without adding more complexity than the team can withstand. Focusing on splitting the `dev` module first will help us iron out the process and ensure that the resulting workflows are sensible before applying them to production in a future migration.
+We feel the [proposed approach][guide-module-splitting] adds a lot of flexibility without adding more complexity than the team can withstand. Splitting by team will allow us to manage Terraform (and Scalr) permissions at the team level, which will be significantly easier than splitting in other ways. Focusing on splitting the `dev` module first will help us iron out the process and ensure that the resulting workflows are sensible before applying them to production in a future migration.
 
-One alternative approach would be to split modules by individual application instead of by team, but that would add complexity without adding much more in the way of flexibility.
+One alternative approach would be to split modules by individual application instead of by team, but that would add complexity without adding much more in the way of flexibility, and is likely more granularity than we need at this point.
 
 ## Alternatives to Module Testing Workflow
 [alternatives-module-testing]: #alternatives-to-module-testing-workflow
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -82,7 +82,7 @@ To address the scaling issues and pain points called out in the [summary][summar
 The following subsections detail each improvement broadly. For specific implementation details, see the [reference-level explanation][reference-level-explanation] section below.
 
 ## 1. Smaller, More Focused Modules
-[guide-module-splitting]: #smaller-more-focused-modules
+[guide-module-splitting]: #1-smaller-more-focused-modules
 
 _Too Many Cooks &rarr; Multiple Kitchens_
 
@@ -99,7 +99,7 @@ A sensible and straightforward solution is to split development resources by tea
 For more details on how this will be accomplished, see the corresponding [reference-level explanation][reference-module-splitting].
 
 ## 2. Testing Workflow
-[guide-module-testing]: #testing-workflow
+[guide-module-testing]: #2-testing-workflow
 
 _Test in isolation_
 
@@ -120,7 +120,7 @@ To truly enforce the "isolation" requirement, module testing will be done in a f
 For details on the proposed testing process, see the corresponding [reference-level explanation][reference-module-testing].
 
 ## 3. Standardize Components
-[guide-standardization]: #standardize-components
+[guide-standardization]: #3-standardize-components
 
 _Don't repeat yourself_
 
@@ -143,7 +143,7 @@ Relatedly, the `ctd-ecs-app` module introduced the concept of a `context` var th
 See the corresponding [reference-level explanation][reference-standardization] for related recommendations.
 
 ## 4. Version and Publish Modules
-[guide-module-versioning]: #version-and-publish-modules
+[guide-module-versioning]: #4-version-and-publish-modules
 
 _Modules as immutable artifacts_
 
@@ -169,7 +169,7 @@ In summary: We should prioritize setting up a versioning and publishing workflow
 For details on the module versioning and publishing workflow, see the corresponding [reference-level explanation][reference-module-versioning].
 
 ## 5. Terraform Automation
-[guide-automation]: #terraform-automation
+[guide-automation]: #5-terraform-automation
 
 _Automate as much as possible_
 
@@ -192,7 +192,7 @@ Note that CTD's ability to procure Scalr is not yet confirmed for FY23, so there
 For details on how Scalr will be configured, and the proposed CI/CD workflows, see the corresponding [reference-level explanation][reference-automation]. There are also some caveats outlined in the [drawbacks section][drawbacks-scalr].
 
 ## 6. User Permissions
-[guide-permissions]: #user-permissions
+[guide-permissions]: #6-user-permissions
 
 _Strike a better balance_
 
@@ -219,7 +219,7 @@ the detailed proposal makes those examples work.
 -->
 
 ## 1. Splitting Root Modules By Team
-[reference-module-splitting]: #splitting-root-modules-by-team
+[reference-module-splitting]: #1-splitting-root-modules-by-team
 
 To optimize the management of development and staging resources, the `dev` root module will be split into multiple root modules that are organized by team. Once a new root module is set up for each engineering team, we'll need to:
 
@@ -277,7 +277,7 @@ Possible future modules:
 - `splunk`
 
 ## 2. Module Testing, Detailed
-[reference-module-testing]: #module-testing-detailed
+[reference-module-testing]: #2-module-testing-detailed
 
 As noted in the [guide-level explanation][guide-module-testing], we will set up testing for each application and infrastructure component child module by creating a `test/` module within each module's directory. The test module will define any dependent resources required by the module, and then call the module itself. It could even call other application modules if needed.
 
@@ -307,7 +307,7 @@ In order to be able to test ECS applications, we will need to figure out how to 
 As noted in the [guide-level explanation][guide-module-testing], testing will happen in the separate `mbta-ctd-test` AWS account. This account will be accessed using "assume role" permissions that full access to manage required AWS resources. The assumed role won't have full administrator access, but we'll manage the list of permitted services via IAM policies.
 
 ## 3. Standardizing on Infrastructure Component Modules
-[reference-standardization]: #standardizing-on-infrastructure-component-modules
+[reference-standardization]: #3-standardizing-on-infrastructure-component-modules
 
 As noted in the [guide-level explanation][guide-standardization], this is currently the most mature of the proposed improvements, in that the `ctd-ecs-app` and `ctd-rds-db` modules already exist and are seeing use. Aside from increasing adoption and carrying out the [testing workflow][guide-module-testing] and [module versioning][guide-module-versioning] recommendations, other improvements that will be beneficial include:
 
@@ -316,7 +316,7 @@ As noted in the [guide-level explanation][guide-standardization], this is curren
 - Standardizing on the use of the `context` variable
 
 ## 4. Module Versioning/Publishing
-[reference-module-versioning]: #module-versioning-publishing
+[reference-module-versioning]: #4-module-versioningpublishing
 
 As noted in the [guide-level explanation][guide-module-versioning], module versioning and publishing is something we will set up for all infrastructure component modules, but which should be considered optional for application modules. See also the [corresponding subsection under rationale and alternatives][alternatives-module-versioning].
 
@@ -347,7 +347,7 @@ Unfortunately this approach is not as advantageous as it seems. Some of the draw
 Given the drawbacks, storing application modules in application repos is not recommended at this time.
 
 ## 5. Automating Terraform
-[reference-automation]: #automating-terraform
+[reference-automation]: #5-automating-terraform
 
 As noted in the [guide-level explanation][guide-automation], our goal for automating Terraform is to procure Scalr. But in order to make optimal use of Scalr for Terraform automation, we need to understand where it will fit into our proposed workflows, and how best to configure it to suit our needs.
 
@@ -411,7 +411,7 @@ Once a new version of the module is published, the module version can be updated
 If module versioning is not employed, changes to the module will need to be applied immediately in any dependent root modules. In the development/staging context, this can be handled automatically by configuring [run triggers](https://docs.scalr.com/en/latest/workspaces.html#run-triggers) in the Scalr workspace. In the production context, however, there will be an extra layer of protection to ensure that changes don't impact production before they're ready to be applied there. This is described in more detail in the [production workflow section][scalr-workflow-production] below.
 
 #### Scalr Workflow for Applying Changes In Development/Staging Context
-[scalr-workflow-development-staging]: #scalr-workflow-for-applying-changes-in-development-staging-context
+[scalr-workflow-development-staging]: #scalr-workflow-for-applying-changes-in-developmentstaging-context
 
 Each root module will have a GitHub-integrated Scalr workspace. Scalr will automatically run `terraform plan` when a pull request is opened, and attempt to automatically run `terraform apply` on merge. It will also be possible to run `terraform plan` locally using Scalr's remote operation backend, but not `terraform apply`; this is to ensure that GitHub acts as the source of truth for controlling Terraform state.
 
@@ -484,7 +484,7 @@ The code owners for each application can include engineering team leads, team me
 In order to ensure that sensitive changes&mdash;such as those which affect IAM resources&mdash;can't be made without being properly vetted, we'll define OPA policies in Scalr that will check for changes to sensitive resource types. For more detail on this topic, see the [Balancing User Permissions][reference-permissions] section below.
 
 ## 6. Balancing User Permissions
-[reference-permissions]: #balancing-user-permissions
+[reference-permissions]: #6-balancing-user-permissions
 
 As noted in the [guide-level explanation][guide-permissions], our goal is to achieve an optimal balance of security and usability with Terraform and AWS. Achieving this goal effectively is somewhat dependent on our ability to delegate permissions to an automated system like Scalr.
 
--- a/rfcs/0000-terraform-workflow-improvements.md
+++ b/rfcs/0000-terraform-workflow-improvements.md
@@ -261,8 +261,12 @@ Root modules associated with the "CTD Production" AWS account that should be ren
     - `aws-ctd-main-team-trc`
     - `aws-ctd-main-team-website`
   - move base dev resources to `aws-ctd-main-base`
-- `onprem_dev` → `aws-onprem-dev`
-- `onprem_prod` → `aws-onprem-prod`
+- `onprem_dev` → multiple modules:
+  - move SSM resources to `aws-ctd-main-base`
+  - move applications to team-based modules
+- `onprem_prod` → multiple modules:
+  - move SSM resources to `aws-ctd-main-base`
+  - move applications to `aws-ctd-main-apps` for now
 - `prod` → `aws-ctd-main-apps` (to remain as a single module for now)
 - `restricted` → `aws-ctd-main-restricted`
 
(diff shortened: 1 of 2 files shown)


rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+This workflow is intended to be carried out manually at first, but we might be able to use GitHub Actions to automate it at some point.
+
+#### Caveats for testing ECS applications
+
+In order to be able to test ECS applications, we will need to figure out how to allow the ECS service to retrieve a container image for deployment. There are a few blockers to this that will need to be overcome:
+
+- We'll need to override the ECR repo URL with the production AWS account's repo, and give the test account permission to pull images
+- We'll need to address an issue where apps can't start until the first GitHub Actions deploy because the template task definition references `latest`:
+  - Option 1: Add `latest` tag to all ECR images
```
boringcactus: It might be more robust to only use the `latest` tag for automatic deploys of `main` to staging, or even for deploys to production, so that experimental code changes don't get accidentally picked up by the Terraform test environment.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+### Root Module Naming Convention
+[root-module-naming-convention]: #root-module-naming-convention
+
+We'll also need to implement a new naming convention for root modules to keep them organized.
+
+Terraform's official [naming convention for published modules][terraform-module-publishing] is `terraform-<PROVIDER>-<NAME>`, where `<NAME>` can contain additional hyphens. Since our own root modules won't be published to Terraform's public registry, we can omit the `terraform-` prefix and use `<PROVIDER>-<NAME>` as a base convention, where `<NAME>` can contain any of the following:
+
+- AWS account nickname (e.g. `mbta-ctd-test`, `mbta-ctd-main`, or preferably a more succinct version that eliminates the `mbta-` prefix)
+- Team name (prefixed with `team-`)
+- Environment context (e.g. `dev`, `staging`, `prod`) - this overlaps with AWS account nickname, so we may want to discourage the use of both at the same time
```
boringcactus: It seems to me like the only overlap is `test` and `ctd-test`, since `dev` and `staging` and `prod` all live under `ctd-main`. Am I missing something?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+- Update references to dependent resources that remain in other modules to use data sources or [remote state](https://developer.hashicorp.com/terraform/language/state/remote-state-data)
+
+Note that production resources will remain in a single `prod` root module in the short term, but will likely eventually also be moved to team-based modules a later date. This will allow us to vet the changes proposed in this document, while also making sure we're adhering to the stricter security requirements for production that are discussed in the [Scalr workflow for production][scalr-workflow-production] section.
+
+### Root Module Naming Convention
+[root-module-naming-convention]: #root-module-naming-convention
+
+We'll also need to implement a new naming convention for root modules to keep them organized.
+
+Terraform's official [naming convention for published modules][terraform-module-publishing] is `terraform-<PROVIDER>-<NAME>`, where `<NAME>` can contain additional hyphens. Since our own root modules won't be published to Terraform's public registry, we can omit the `terraform-` prefix and use `<PROVIDER>-<NAME>` as a base convention, where `<NAME>` can contain any of the following:
```
boringcactus: [RFC 13](https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0013-consistent-resource-names.md) specifies that Terraform names should be `snake_case`. If modules are an exception, they should probably be documented there as an exception.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+  - It's required if publishing to the public Terraform registry
+- Ensure that the repo name follows the convention `terraform-aws-<module-name>`, which is a requirement for both [Terraform's public registry][terraform-module-publishing] and [Scalr's private registry][scalr-private-registry]
+- Set up a tagging/release workflow in GitHub Actions that publishes the module to Scalr's private module registry or the public Terraform registry (ideally as a reusable workflow)
+- Configure any references in root modules to call the module from the repo or registry instead of a relative path to the module
+
+### Application Modules In Application Repos?
+
+For application modules, the question has been raised as to whether it might make sense to move the module configuration into the application repo so it lives alongside the codebase. The stated advantages to this approach are as follows:
+
+- All(?) of the application's configuration would live within the same repo
```
boringcactus: An advantage I'd been thinking of to storing the config alongside the code is that deploying the code could also apply the config, so that the code would always see, for example, the environment variables it was expecting. I still think this would be good, but it would introduce some new challenges, and it seems like it'd require a complete overhaul of the entire structure of our Terraform setup to get to a point where that's even possible.

If the config and the code are deployed independently, then they should be versioned independently, so I agree that it makes more sense for them to be stored independently.

boringcactus (approved):
I'm excited.

Notes on the unresolved questions:
- The proposed module names mostly make sense to me. I'm not sure how the `onprem` in `aws-onprem-dev` fits into the naming convention, and I'm not sure what makes `aws-ctd-main-apps` a better name than `aws-ctd-main-prod` (since environment context is listed before component descriptor and also since we have apps in the team-specific modules too).
- I'm not familiar with the status quo of Terraform testing, so I'm not sure when the module testing workflow would need to be used, or what a test case would look like. Maybe the guide-level explanation is still too specific, or maybe that whole topic isn't relevant to engineers on application teams, or maybe something else, I'm not sure.
- I think standardizing on the same boilerplate in `ctd-ecs-app` and `ctd-rds-db` will be helpful, and I think the infrastructure team are better situated than engineers from application teams to do that refactoring, especially when the shared modules will need to be extended to accommodate new behaviors they can't yet express.
- My perception is that our open-source applications mostly don't have users outside the T anyway, and so public Terraform modules for them would be basically irrelevant. That said, if there's no difference in the amount of work it takes to make them public vs private, we might as well make them public, on the same grounds as the code itself and also this repository right here.
- I think we don't need to rush to start versioning application modules until and unless we run into problems caused by a lack of versioning for application modules.
- I think pinning prod modules by Git hash but keeping prod as one module makes sense as a short-term approach.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+#### Proposed Module Names
+
+With all the above in mind, the following are the proposed names for all existing and future root modules. (Note that the naming convention for child modules is different, and is covered in the [module versioning][guide-module-versioning] section.)
+
+Root modules associated with the "CTD Production" AWS account that should be renamed:
+
+- `bootstrap` → `aws-ctd-main-bootstrap`
+- `global` → `aws-ctd-main-base`
+- `dev` → multiple modules:
+  - move application development/staging resources to team-based modules:
```
skyqrose: Is team-based the best way to split these? Another way could be to put resources together if they depend on each other / reference each other, which could lead to smaller modules if one team manages multiple unrelated resources. (I'm thinking in particular, the TRC team.)

I'm curious if there's a list of all the current tf files in dev, which ones need to be in the same module for technical/dependency reasons, which ones need to be in the same module for team/permission/administrative reasons, and which ones would end up in each of these team modules as currently proposed.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+#### Proposed Module Names
+
+With all the above in mind, the following are the proposed names for all existing and future root modules. (Note that the naming convention for child modules is different, and is covered in the [module versioning][guide-module-versioning] section.)
+
+Root modules associated with the "CTD Production" AWS account that should be renamed:
+
+- `bootstrap` → `aws-ctd-main-bootstrap`
+- `global` → `aws-ctd-main-base`
+- `dev` → multiple modules:
```
skyqrose: Will any tf files be split or merged, or will this only operate on whole tf files?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+Possible future root modules:
+
+- `github`
+- `sentry`
+- `splunk`
+
+## 2. Module Testing, Detailed
+[reference-module-testing]: #2-module-testing-detailed
+
+As noted in the [guide-level explanation][guide-module-testing], we will set up testing for each application and infrastructure component child module by creating a `test/` module within each module's directory. The test module will define any dependent resources required by the module, and then call the module itself. It could even call other application modules if needed.
```
skyqrose: So, if I understand this correctly, the directory structure would look like this:
- `terraform/aws-mbta-ctd-main-team-myapp/`
  - `main.tf` (define the module)
  - `staging.tf` (instantiate the module in mbta-ctd-**main**)
  -  `dev-green.tf` (instantiate the module in mbta-ctd-**main**)
  - `test/`
    - `test.tf` (instantiate the module in mbta-ctd-**test**)
    
This seems to slightly break the naming convention, in that a file in `*mbta-ctd-main*/` instantiates resources in a different aws account. That's probably fine though, it makes sense.

But it does make me a little worried about the potential for error when running the tests. If you forget to `cd test` when running `terraform apply`, you'd then apply to the wrong aws account, and that seems like an easy thing to overlook.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+Relatedly, the `ctd-ecs-app` module introduced the concept of a `context` var that bundles all baseline configuration need by a module, including VPC and subnet information, and shared resources such as TLS certificates and Cognito domains. This convention has the potential to significantly reduce redundant configuration in app modules, and should also be adopted widely.
+
+See the corresponding [reference-level explanation][reference-standardization] for related recommendations.
+
+## 4. Version and Publish Modules
+[guide-module-versioning]: #4-version-and-publish-modules
+
+_Modules as immutable artifacts_
+
+Nearly all of CTD's modules are currently stored in the devops repo and called using a relative path to the module's directory. Terraform supports a number of alternate methods for sourcing modules, including:
```
skyqrose: Should `terraform/` be its own github repo? The `terraform/` directory in the `devops` repo kind of overshadows everything else there, and it seems like it's big enough and freestanding enough to be its own repo.

Pros:
- Single-purpose repos are neater than giant bags of miscellany. Less mess in the git history, less files to work around, shorter paths, etc.
- Permissions can be more tightly controlled. I don't need write access to the repo with ansible config, but I do need write access to the repo with terraform.

Cons:
- Migrating to a new dedicated repo would be a royal pain
- Infra resource management would be split between more places for people who do use all of it.

If we did want to pull it into a new repo, this seems like the time and place to figure that out.

Alternative: Moving modules that belong to a single application into the application's repo. Is that under consideration?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+## 3. Standardizing on Infrastructure Component Modules
+[reference-standardization]: #3-standardizing-on-infrastructure-component-modules
+
+As noted in the [guide-level explanation][guide-standardization], this is currently the most mature of the proposed improvements, in that the `ctd-ecs-app` and `ctd-rds-db` modules already exist and are seeing use. Aside from increasing adoption and carrying out the [testing workflow][guide-module-testing] and [module versioning][guide-module-versioning] recommendations, other improvements that will be beneficial include:
+
+- Handling all use cases for each module, using vars to enable/disable features that aren't likely to be used everywhere
+- Building in support for maintenance modes to minimize temporary configuration changes
+- Standardizing on the use of the `context` variable
```
skyqrose: The `context` is currently defined in `terraform/dev/main.tf` and `terraform/prod/main.tf`. How would it be defined in a global way if resources are broken into many more modules?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+### Application Modules In Application Repos?
+
+For application modules, the question has been raised as to whether it might make sense to move the module configuration into the application repo so it lives alongside the codebase. The stated advantages to this approach are as follows:
+
+- All(?) of the application's configuration would live within the same repo
+- Open source repos would include the necessary infrastructure configuration to deploy the application
+
+Unfortunately this approach is not as advantageous as it seems. Some of the drawbacks include:
+
+- Configuration that varies by environment will still live in root modules in the devops repo, so there will always be some application configuration that lives outside the application repo
```
skyqrose: Does this mean the current `dev/myapp-dev.tf` file? Why can't that belong alongside the application module in the application repo?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+As noted in the [guide-level explanation][guide-standardization], this is currently the most mature of the proposed improvements, in that the `ctd-ecs-app` and `ctd-rds-db` modules already exist and are seeing use. Aside from increasing adoption and carrying out the [testing workflow][guide-module-testing] and [module versioning][guide-module-versioning] recommendations, other improvements that will be beneficial include:
+
+- Handling all use cases for each module, using vars to enable/disable features that aren't likely to be used everywhere
+- Building in support for maintenance modes to minimize temporary configuration changes
+- Standardizing on the use of the `context` variable
+
+## 4. Module Versioning/Publishing
+[reference-module-versioning]: #4-module-versioningpublishing
+
+As noted in the [guide-level explanation][guide-module-versioning], module versioning and publishing is something we will set up for all infrastructure component modules, but which should be considered optional for application modules. See also the [corresponding subsection under rationale and alternatives][alternatives-module-versioning].
```
skyqrose: Are there any situations where a publishing/versioning an application module would make sense? Generally, they just have a couple instantiations for a couple environments, all managed by the same team.

If this RFC changed to "won't do this for application modules", then the tradeoffs become easier to analyze, and this wouldn't conflict with other discussions about how to manage application modules, like moving them into the application repo.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+  before stabilization?
+- What related issues do you consider out of scope for this RFC that could be addressed in the
+  future independently of the solution that comes out of this RFC?
+-->
+
+One major unresolved question is the timeline for implementing all the improvements laid out in this RFC. As noted in the [Drawbacks][drawbacks] section above, it's a lot of work. Once this RFC is approved we'll need to scope out the work and divide it accordingly between the infrastructure and engineering teams.
+
+Other questions for the reader:
+
+- Is the proposed [root module naming convention][root-module-naming-convention] sensible and sufficiently comprehensive?
```
skyqrose: It could use an example, real-world or fake.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+- Does the [isolated module testing workflow][reference-module-testing] make sense? Does it seem easy enough to carry out?
```
skyqrose: One thing I think I missed in it is how do you know if it worked? If it completed without errors? Are there health checks or assertions to make? Would it be usable for application end-to-end tests?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+- How aggressive should we be about [standardizing on infrastructure component modules][guide-standardization] for ECS and RDS across all applications/teams?
+- How aggressive should we be about [versioning and publishing application modules][guide-module-versioning] for open-source applications?
```
skyqrose: Not at all. If anybody else wants to use an open-source application, they're probably going to have a completely different infrastructure and deployment setup than us. It's perhaps most useful as a reference of what a working deployment is like, which env vars are filled in which ways, but I wouldn't expect anyone to take a module for an open source repo and deploy it as is.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+- Branch protection rules ensure that PR cannot be merged without approval
+
+##### Applying Changes
+
+- Engineer merges approved PR
+- If module versioning is employed, GitHub Actions triggers workflow to publish new version of module to module registry
+
+Once a new version of the module is published, the module version can be updated in the development/staging and production contexts using the workflows described below.
+
+If module versioning is not employed, changes to the module will need to be applied immediately in any dependent root modules. In the development/staging context, this can be handled automatically by configuring [run triggers](https://docs.scalr.com/en/latest/workspaces.html#run-triggers) in the Scalr workspace. In the production context, however, there will be an extra layer of protection to ensure that changes don't impact production before they're ready to be applied there. This is described in more detail in the [production workflow section][scalr-workflow-production] below.
```
skyqrose: When would one module depend on another application module, and not be in the same terraform root module?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+Other questions for the reader:
+
+- Is the proposed [root module naming convention][root-module-naming-convention] sensible and sufficiently comprehensive?
+- Does the [isolated module testing workflow][reference-module-testing] make sense? Does it seem easy enough to carry out?
+- How aggressive should we be about [standardizing on infrastructure component modules][guide-standardization] for ECS and RDS across all applications/teams?
+- How aggressive should we be about [versioning and publishing application modules][guide-module-versioning] for open-source applications?
+- If changes to non-versioned application modules can spill over into other modules, as noted in the [Scalr Workflow for Module Testing][scalr-workflow-module-testing], is that enough of a reason to push to version and publish all application modules?
+- Does the decision to [delay splitting the `prod` module and instead version all production module calls with a git hash][scalr-workflow-production] make sense?
```
skyqrose: Switching to using git hashes seems like a bit of a headache that has potential to go wrong, and it'd be nice to avoid thinking about versioning for application modules. Continuing without them is no different from the status quo, where we make and protect changes to application modules with the understanding that they affect prod.

If we do want to eventually move prod config into the application root module, then I don't think adding that to the scope adds too much risk to this RFC, and we should plan to do it rather than delay it indefinitely.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how programmers should *think* about the feature, and how it should impact the way they
+  work on this project. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to senior developers and to junior
+  developers.
+-->
+
+To address the scaling issues and pain points called out in the [summary][summary] and [motivation][motivation] sections, the following improvements are recommended:
```
skyqrose: What order will these 6 steps be done in?

skyqrose (commented):
Hooray! I've been hoping for something like this.

The biggest confusion point for me was about versioning application modules. I'm still not sure what problem it's solving, and it seems to add a lot of complexity.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+Thankfully, the majority of the infrastructure maintained by engineering teams is defined in independent _modules_ (also sometimes referred to as _child modules_ to differentiate them from the _root modules_ that directly manage infrastructure state). Each application's module defines a set of infrastructure components, either by defining Terraform resources directly, or calling other modules that define infrastructure components.
+
+The use of modules is a Terraform best practice; modules allow engineers to define all of an application's infrastructure as a self-contained set of configuration, which can then be used to create multiple nearly identical environments. Since modules are intended to be largely independent and reusable, it's relatively easy to create and destroy a module's resources separately from the context where the module will ultimately be deployed.
+
+In short: **To test our infrastructure configuration in Terraform, we should focus on testing modules.**
```
mzappitello: 💯

rfcs/0000-terraform-workflow-improvements.md
mzappitello: there is a part of me that finds the concept of moving the module to our repo particularly appealing. having the infrastructure described in the same place as the code makes me feel like testing and deployments could be done a bit cleaner, especially in the case where we make a change across both of the, like defining an env var. it also _feels nice_ that everything needed to describe the project exists in one repo.

i'm not sure either of those are compelling enough reasons and i'm sure its a pain to switch to, but there is something about this approach that my mind keeps returning to.

rfcs/0000-terraform-workflow-improvements.md
mzappitello: i would think that you have something more like

```
terraform/aws-mbta-main-team/
    variables.tf
    outputs.tf
    main.tf
    staging/
        variables.tf
        staging.tf
    dev/
        variables.tf
        dev.tf
        dev-blue.tf
    test/
        variables.tf
        main.tf
```

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+To use a module's `test/` module for testing:
+
+1. On a new feature branch, make modifications to the functionality of the module
+1. In that same branch, add test cases for the new functionality to the module's testing configuration in `test/`
+1. Switch to the default branch (`master`/`main`), `cd` to the `test/` directory and run `terraform apply` to stand up the test resources
+1. Switch to the feature branch and run `terraform plan`/`terraform apply` to see the impact of the changes on the pre-existing resources
+1. When finished testing, run `terraform destroy` to tear down all test resources
+
+This workflow is intended to be carried out manually at first, but we might be able to use GitHub Actions to automate it at some point.
```
mzappitello: when using this workflow i mostly enjoyed how quickly i could get feedback on the changes i wanted to make. one thing to note though is that standing up databases can take quite a bit of time. it wasn't an end of the world problem, but if a project uses an RDS and is standing up multiple tests, the automated tests could stretch out.

rfcs/0000-terraform-workflow-improvements.md
mzappitello: not sure if this is useful others, but a workflow where you provide a git hash for the application repo that will run deploy scripts on that tag and then pull that image for the newly instantiated `dev` or `staging` environment would be useful for me.

rfcs/0000-terraform-workflow-improvements.md
mzappitello: we have a pretty big need for them on lamp.

 a good portion of the work we do is infrastructure R&D as we try to find the right combinations of infra and code solutions. we could probably achieve some of this with feature flags, but we could end up with a bunch of them, especially if we just want to run a test. versioning and only deploying versioned modules to production is our preferred setup.
 
i'm pretty sure (or as sure as i can be given i don't know most of the details) that versioning the ctd shared modules (like for RDS and ECS) is an important feature to work out so that teams can update their versions on their own schedule if breaking changes need to be applied.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+## 4. Module Versioning/Publishing
+[reference-module-versioning]: #4-module-versioningpublishing
+
+As noted in the [guide-level explanation][guide-module-versioning], module versioning and publishing is something we will set up for all infrastructure component modules, but which should be considered optional for application modules. See also the [corresponding subsection under rationale and alternatives][alternatives-module-versioning].
+
+### How To Publish A Module
+
+To set up a module for versioning and publishing, we'll need to:
+
+- Move the module's configuration to a separate repo, for a few reasons:
```
mzappitello, line 333: can (should?) that separate repo be the one that contains the code that is run on that infrastructure?

rfcs/0000-terraform-workflow-improvements.md
mzappitello: it seems like @ianwestcott addresses this later on in this document.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+
+If module versioning is not employed, changes to the module will need to be applied immediately in any dependent root modules. In the development/staging context, this can be handled automatically by configuring [run triggers](https://docs.scalr.com/en/latest/workspaces.html#run-triggers) in the Scalr workspace. In the production context, however, there will be an extra layer of protection to ensure that changes don't impact production before they're ready to be applied there. This is described in more detail in the [production workflow section][scalr-workflow-production] below.
+
+#### Scalr Workflow for Applying Changes In Development/Staging Context
+[scalr-workflow-development-staging]: #scalr-workflow-for-applying-changes-in-developmentstaging-context
+
+Each root module will have a GitHub-integrated Scalr workspace. Scalr will automatically run `terraform plan` when a pull request is opened, and attempt to automatically run `terraform apply` on merge. It will also be possible to run `terraform plan` locally using Scalr's remote operation backend, but not `terraform apply`; this is to ensure that GitHub acts as the source of truth for controlling Terraform state.
```
mzappitello: is this a requirement for dev environments as well? having to go through pull requests to run an apply feels a bit cumbersome to me.

ianwestcott:
@boringcactus thanks! Some responses to your main feedback (the rest I'll address inline as needed)

> I'm not sure how the `onprem` in `aws-onprem-dev` fits into the naming convention

That's an odd root module &ndash; it manages on-premise infrastructure, but it does so via [AWS Systems Manager (SSM)](https://aws.amazon.com/systems-manager/). The resources being managed by that module are SSM resources, hence the `aws-` prefix. But now that I'm scrutinizing that module, it might make more sense to fold it into other modules: `aws-ctd-main-base` for the SSM resources and team-based modules for the applications.

> I'm not sure what makes `aws-ctd-main-apps` a better name than `aws-ctd-main-prod` (since environment context is listed before component descriptor and also since we have apps in the team-specific modules too).

Good point &ndash; I was trying to steer away from environment context because those words ("dev"/"prod") are a bit overloaded. But in this case it may be the best term available. And whenever we get around to moving production applications into team-based modules, this module will ultimately be emptied out and removed, so maybe it won't hurt to keep at least part of its current name in the meantime.

> I'm not familiar with the status quo of Terraform testing, so I'm not sure when the module testing workflow would need to be used, or what a test case would look like. Maybe the guide-level explanation is still too specific, or maybe that whole topic isn't relevant to engineers on application teams, or maybe something else, I'm not sure.

My goal is to make it easy for every engineer that works with Terraform to be able to test the modules they maintain without having to perform that testing against the infrastructure running in our main AWS account. I believe that teams will benefit from being able to test their application modules in isolation in a similar manner to how they test their application code. Such testing may not need to be carried out every day, but it would be useful whenever there are any major changes to the application being proposed, such as adding or removing infrastructure components, or making a change that requires replacing a resource.

Does that add enough clarity or is something still missing? I can update the guide-level explanation accordingly.

> I think standardizing on the same boilerplate in `ctd-ecs-app` and `ctd-rds-db` will be helpful, and I think the infrastructure team are better situated than engineers from application teams to do that refactoring, especially when the shared modules will need to be extended to accommodate new behaviors they can't yet express.

Noted, though unfortunately this may be less a function of expertise than of available resources.

> My perception is that our open-source applications mostly don't have users outside the T anyway, and so public Terraform modules for them would be basically irrelevant. That said, if there's no difference in the amount of work it takes to make them public vs private, we might as well make them public, on the same grounds as the code itself and also this repository right here.

Public Terraform modules [are indeed more work](https://developer.hashicorp.com/terraform/registry/modules/publish), in part because they will need to be in separate repos in order to be registered, and will need to follow a [more rigorous module structure](https://developer.hashicorp.com/terraform/language/modules/develop/structure). The question of relevance likely varies by application, so this is something that I think is best handled on a case-by-case basis.

> I think we don't need to rush to start versioning application modules until and unless we run into problems caused by a lack of versioning for application modules.

I think we trip over the [drawbacks to the current approach](https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#alternative-1-do-nothing) fairly often, but maybe the frequency varies by team. 

> I think pinning prod modules by Git hash but keeping prod as one module makes sense as a short-term approach.

:+1:

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I think what's missing is that we were originally thinking of changing the main AWS account's name to `mbta-ctd-prod` when this was written, but we later settled on `mbta-ctd-main` to solve exactly this problem. I'll remove this note.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: That's true of resource names but not module names. I'll make a task to update that document to clarify the distinction.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I touched on this briefly [here](https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#alternatives-to-splitting-root-modules):

> One alternative approach would be to split modules by individual application instead of by team, but that would add complexity without adding much more in the way of flexibility.

For clarity, I'll expand that section to add that splitting by team allows us to manage Terraform (and Scalr) permissions at the team level, which I think is going to be a lot easier than splitting in other ways. I also worry that a team managing multiple smaller modules might be more granularity that we need at this point, though @bklebe is welcome to disagree with me on that. 😉 

Most of our application modules don't reference each other, and those that do could use [output values](https://developer.hashicorp.com/terraform/language/values/outputs), so there's no technical limitation here. There might be dependency issues, which we'll likely handle by prioritizing certain applications/teams higher in our disaster recovery plan.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I'm pretty sure whole files only, since currently nearly all files within the `dev` root module represent one application each. There are a couple small exceptions for shared resources, which will be moved to `aws-ctd-main-base` and referenced via output values.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: The module testing workflow wouldn't apply to root modules, only child modules. Since your `aws-mbta-main-team` module is a root module, it won't contain any subdirectories, only `.tf` files that reference child modules such as `app-something`. It's your `app-something` child module that would contain its own testing module within it. All testing would be on the individual child module, and changes would then bubble up to the root module once tested, merged (and optionally versioned).

Does that make sense?

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: @skyqrose:

> Should terraform/ be its own github repo?

Maybe 😅 

> Migrating to a new dedicated repo would be a royal pain

I agree, which is the main reason why I'm not advocating for it. But I could be convinced if there's enough interest.

> Alternative: Moving modules that belong to a single application into the application's repo. Is that under consideration?

Discussed here: https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#application-modules-in-application-repos

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I think there would still be a `context` var defined in a `locals` block in each module, but its values could come from either [output values](https://developer.hashicorp.com/terraform/language/values/outputs) of other modules, or a [`.tfvars` file](https://developer.hashicorp.com/terraform/language/values/variables#variable-definitions-tfvars-files) that we populate in some other manner. (Perhaps symlinked from elsewhere in the repo?)

I'll think more on this and update accordingly.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: You mean like a separate root module for each application, instead of for each team? Let me sit with this question, because I want to make sure I can give a thoughtful answer.

As @boringcactus' [noted above](https://github.com/mbta/technology-docs/pull/16#discussion_r1024642240), it would mean a lot of fundamental changes to our Terraform workflow that aren't currently reflected in this RFC.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: Hm, does the [Proposed Module Names](https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#proposed-module-names) section not satisfy this need?

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: This might be context-dependent, but I think "completed without errors" will be a good baseline for most modules. Assertions are an interesting idea, though I'm not sure how we'd carry that out exclusively in Terraform. ([this third-party provider might be an option.](https://registry.terraform.io/providers/bwoznicki/assert/latest/docs))

I do think end-to-end tests are something that _could_ be possible, but I don't want to go too deep in recommending strategies that are specific to a certain type of application.

I'll see if I can add more to the workflow doc to clarify this.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I'm not sure I understand this question, but maybe it helps to clarify that the module testing workflow will only apply to _child_ modules, such as application modules and infrastructure component modules. Let me know if I should make that clearer in the text.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: > Switching to using git hashes seems like a bit of a headache that has potential to go wrong

I agree that it will add some overhead, but in what ways do you think it could go wrong?

> Continuing without them is no different from the status quo, where we make and protect changes to application modules with the understanding that they affect prod.

My concern is that the concurrent work to [add automation around Terraform](https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#5-terraform-automation) will limit this understanding and increase the chance that module changes unintentionally impact production, which is a significant risk. I'm hoping we can strike a balance that requires less infrastructure team oversight on small changes while still ensuring that every change is deliberate and has a known impact. Leaving less to intuition is worth the additional step, IMO.

> If we do want to eventually move prod config into the application root module, then I don't think adding that to the scope adds too much risk to this RFC, and we should plan to do it rather than delay it indefinitely.

Fair point. I want to make sure we have the process and automation figured out before we take on production, but that doesn't have to mean calling all of production out of scope. I'll think more about this.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: That remains to be determined, but we're planning a scoping session soon where we'll iron out dependencies. Some parts may allow (or even require) other parts to be done concurrently.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: This is a technical limitation of GitHub-integrated Scalr workspaces, so it will apply to both dev and prod. That said, the expectation is that any changes will have to be reviewed before they're applied, which aligns with our current engineering and devops workflows. Let me see if I can make that more explicit.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,826 @@
...
+- `global` → `aws-ctd-main-base`
+- `dev` → multiple modules:
+  - move application development/staging resources to team-based modules:
+    - `aws-ctd-main-team-dataplatform`
+    - `aws-ctd-main-team-glides`
+    - `aws-ctd-main-team-lamp`
+    - `aws-ctd-main-team-pass-programs`
+    - `aws-ctd-main-team-screens`
+    - `aws-ctd-main-team-trc`
+    - `aws-ctd-main-team-website`
```
thecristen: These names are so verbose. I understand why each part is there except the `main` -- do we anticipate there being `aws-ctd-nonmain-team-*` modules?

rfcs/0000-terraform-workflow-improvements.md
skyqrose: I meant an example of what one of those modules would look like. What's contained within one root module's / team's directory. One application right now has multiple files across the `devops/terraform/` directory tree. (I may be conflating "root module" and "directory" because we currently group things that way.)

My motivation for asking this was that for the sorts of terraform changes I do regularly (application config changes), I want a better idea of what my experience making a change to my app would be like in this naming / organization scheme.

rfcs/0000-terraform-workflow-improvements.md
skyqrose: Yeah, I misunderstood the root module / application module relationship, this question is nonsense, you can ignore it.

rfcs/0000-terraform-workflow-improvements.md
skyqrose: Hypothetically if the root module did move to the application module, then prod would need to be part of the that. So it seems like there's some pretty strict dependency:
1. reorganize the dev modules into teams
2. move the prod modules from the shared prod module into the team modules
3. move a team module into an application repo

And we don't need to (and probably shouldn't) decide whether to do step `n+1` until we've completed step `n`.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: The word `main` is derived from the (future) AWS account nickname `aws-ctd-main`. See explanations [here](https://github.com/mbta/technology-docs/pull/16/files/202157eeb09a72fed26dc01bda0764f1c763b5ef#diff-f30d4f210d145fc9215534dd17062f685f7bc5a6f4522b4ca6b3ef7b0e39f9faR238) and [here](https://github.com/mbta/technology-docs/pull/16/files/202157eeb09a72fed26dc01bda0764f1c763b5ef#diff-f30d4f210d145fc9215534dd17062f685f7bc5a6f4522b4ca6b3ef7b0e39f9faR244).

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: @skyqrose Agreed. I also think that, while there's no technical reason why we couldn't do this, there are some administrative wrinkles which could complicate implementation.

As noted in [this section](https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#scalr-workflow-for-applying-changes-in-production-context), we need to maintain some sort of change approval for infrastructure changes. Separately, we need to be careful about how we automate [management of AWS permissions](https://github.com/mbta/technology-docs/blob/idw-terraform-workflow-improvements/rfcs/0000-terraform-workflow-improvements.md#aws-permissions) for applications. Thinking through both of those problems is in scope for the implementation of this RFC. But I think we need to make sure we arrive at the best solutions for both approvals and permissions changes before I'd feel comfortable breaking root modules out of the monorepo model. And even then, having to scale those solutions horizontally across `n` repos might present its own challenges.

So for the time being I'd prefer to declare that option out of scope, but perhaps it's something we can revisit in the future.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: Added some clarification: https://github.com/mbta/technology-docs/pull/16/commits/de7cdb955da96edb4d5f32ac96d941e07a6ca65f

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: https://github.com/mbta/technology-docs/pull/16/commits/737d6877b0e11a683d750681232709ef007932c2

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: https://github.com/mbta/technology-docs/pull/16/commits/cccc1f12f453d93a52a68b6737e3373d312261b8

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I've updated the RFC to recommend moving modules to a separate monorepo: https://github.com/mbta/technology-docs/pull/16/commits/9af82d2e1a03c69231e720a5112ce072a720f514

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: https://github.com/mbta/technology-docs/pull/16/commits/7478c9312250a5adbd06c2218e95ba866867be59

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I think your day-to-day experience won't change at all, because you'd be working _within_ a module that has a set name per the recommendations in this document. How you structure individual files within your root modules (and even child modules) is up to your team's discretion, though Terraform [does offer recommendations](https://developer.hashicorp.com/terraform/language/modules/develop/structure) for standard filenames such as `main.tf`, `variables.tf`, and `outputs.tf`.

Let me know if that answers your question.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: https://github.com/mbta/technology-docs/pull/16/commits/80af5b572a6dbd4a614fa314f5b37756dd348a5b

ianwestcott:
Some changes to this document that could use additional review/scrutinization:

**Production Infrastructure In Team Root Modules:** Based on feedback, we've decided to _**include the recommendation to move production infrastructure into team-based root modules**_, which was previously deemed out of scope. We'll still need to wait until after development infrastructure is migrated, in order to vet the workflow and make sure the security requirements are met.

**Application Module Versioning:** We've updated the module versioning and publishing section to recommend that _**all application and infrastructure component modules be migrated to a versioned workflow.**_ Rationale in bullet form:

- The original recommendation shied away from versioning all application modules in order to avoid the increased complexity that versioning will introduce
- However, partway through the process of investigating the recommendations, we realized that it’s necessary to explicitly declare what infrastructure runs in production, in order to keep infrastructure secure and enforce change control requirements
- Our previous recommendation to append git hashes to production module calls was a sort of band-aid compromise to address the change control requirement
- In the interim, I've thought through the pros and cons and reconsidered my recommendation
- Instead of defaulting to not versioning and relying on half-baked versioning system for production, it makes more sense to do the opposite: Default to versioning all modules, but make those version constraints optional in development contexts
- Figuring out how to manage the complexity of versioning modules is not an insurmountable obstacle, and the relevant sections of the RFC have been updated to lay out the process for migrating our modules
- We're also recommending that we move application modules to a separate monorepo from route modules, which will isolate the CI configuration, git tags, and commit history for modules
- Finally, we'll make use of Scalr's private registry so that we can take full advantage of Terraform functionality around declaring module versions (e.g. flexible version constraints), which is more robust than using Git tags directly

rfcs/0000-terraform-workflow-improvements.md
skyqrose: The recent decision [to move production infrastructure into team-based root modules](https://github.com/mbta/technology-docs/pull/16#issuecomment-1426098991) affects this.

We're planning to do step 2 above, so now could do step 3, but it still doesn't need to (shouldn't) be decided now, and the other issues in this thread still apply.

This line of the RFC is now a little incorrect though, and if you feel like being thorough you could update it.

ianwestcott: I don't think migrating production infrastructure into team-based root modules necessarily informs whether team-based root modules can/should be moved to application repos. While it's feasible on a technical level, the decision of whether or not to allow root modules outside of a monorepo has administrative and security implications that are more consequential. And as long as:

1. Application modules live in a designated monorepo (as the updated recommendation stipulates)
1. Root modules remain in a separate monorepo (as is currently in scope)

Then I think this line will still be correct.

rfcs/0000-terraform-workflow-improvements.md
skyqrose: Okay, I'm understanding it a bit more. I do think an example would still help make it clearer, and could prevent a little bit of "am I putting this in the right place" when implementing this. As an application developer, I don't think of it as root modules and child modules, I think of it as which files am I editing, so a concrete example would help me summarize and digest all of the complicated parts of this RFC.

Let's say there's a team with a relatively standard structure with 1 or 2 apps that each have prod/staging/dev-green environments, some global resources that aren't per environment (like s3 buckets), and references the shared ctd-app-ecs. What repos/files does the team care about in the new organization, which files are referenced with versions, and which files need tests?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,856 @@
...
+- Building in support for maintenance modes to minimize temporary configuration changes
+- Standardizing on the use of the `context` variable
+
+## 4. Module Versioning/Publishing
+[reference-module-versioning]: #4-module-versioningpublishing
+
+As noted in the [guide-level explanation][guide-module-versioning], module versioning and publishing is a necessary improvement that should be set up for all infrastructure component and application modules. To achieve this, we'll need to make some changes to how our modules are stored and structured. In particular, we'll need to rename module directories to adhere to Terraform's module naming convention of `terraform-<PROVIDER>-<NAME>`.
+
+We'll also need to migrate all child modules to a separate monorepo, which we'll call `terraform_modules`. This will allow us to isolate the CI configuration, git tags, and commit history for modules. It will also help reinforce the separation between child modules and root modules, since it will no longer be possible to update both a child module and the infrastructure that uses it in the same pull request.
```
beatrixada: > It will also help reinforce the separation between child modules and root modules, since it will no longer be possible to update both a child module and the infrastructure that uses it in the same pull request.

I may be misreading but this sounds undesirable? At least in terms of increasing the number of pull requests it takes to change something, which means that they have to be tracked separately and merged in a specific order.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,856 @@
...
+   - Ideally the tagging workflow will be automated and derive the module name and appropriate version string automatically (i.e. from semantic commits or PR tags)
+1. Migrate each infrastructure component and application module to the new repo and rename it to match the required `terraform-aws-<module_name>` convention
+1. Add each module to the private module registry
+1. Update any references in root modules to refer to the module registry, requiring version constraints in production, but not in dev/staging
+
+### Public Vs. Private Registries
+
+There are two options for publishing modules: Using [Terraform's public registry](https://registry.terraform.io/), or using a private registry such as the one [offered by Scalr][scalr-private-registry].
+
+Publishing a module to Terraform's public registry comes with its own [strict set of requirements][terraform-module-publishing], not least of which being that it must be stored in a discrete public repo. Thankfully, private registries is less restrictive on these points, so publishing to a private registry is the preferred approach unless there's a reason to make a module public.
```
beatrixada: typo:
```suggestion
Publishing a module to Terraform's public registry comes with its own [strict set of requirements][terraform-module-publishing], not least of which being that it must be stored in a discrete public repo. Thankfully, private registries are less restrictive on these points, so publishing to a private registry is the preferred approach unless there's a reason to make a module public.
```

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,856 @@
...
+
+If we do choose to publish a module in Terraform's public registry, the overall process will be similar, except that we'll need to make a separate repo for the module and make other changes in accordance with Terraform's requirements.
+
+### Application Modules In Application Repos?
+
+For application modules, the question has been raised as to whether it might make sense to move the module configuration into the application repo so it lives alongside the codebase. The stated advantages to this approach are as follows:
+
+- All(?) of the application's configuration would live within the same repo
```
beatrixada: I concur with @boringcactus and think this is a very underrated advantage, as it means you can have atomic PRs that ensure code isn't broken between merging two different PRs and don't require tracking multiple PRs across repos to make changes.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,856 @@
...
+- Open source repos would include the necessary infrastructure configuration to deploy the application
```
beatrixada: Frankly, I feel more neutral than positive about this. I think we should be prioritizing changes that increase productivity, though I also realize we haven't clearly prioritized our commitment to open source vs productivity.

rfcs/0000-terraform-workflow-improvements.md
beatrixada: I hear you on scope @ianwestcott, but nonetheless I think it's worth taking a more fundamental look at our process here.

In particular, I don't quite understand the administrative/security objections. It seems to me that change approval would still happen in the form of code review on the pull request in the app repo? And as regards permissions, I don't fully understand Scalr's capabilities, but hopefully it should be able to apply different root configurations with different sets of permissions? To me that would be _preferable_ security-wise to having one large root configuration with broad permissions for many applications.

The current process for infrastructure changes can be quite slow and requires managing infra changes out-of-band to application changes, frequently blocking those application changes and reducing deploy frequency. I'm not sure that moving child modules into team-specific repos addresses this problem. Coordinating between infra and app changes currently requires PRs in two separate repos. It seems like after this change, it could require coordinating _three_ separate PRs instead?

I don't expect drastic changes to be adopted at this point in the RFC's lifecycle and I apologize for not raising my concerns sooner, but I think we should continue to improve our process in this area even after making these changes.

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,856 @@
...
+
+More importantly, versioning allows us to explicitly declare what version of an application module is running in the production context, and enforce a change control process to ensure that no unwanted or unintended changes are applied. Change control on production systems is a security requirement, which CTD typically handles via code reviews or deployment approvals. Without module versioning, changes to child modules could have immediate and undetectable impacts to production systems. By pinning modules in production, we can isolate changes to each running environment, and thus isolate the approval step for an extra layer of protection.
```
beatrixada: Re: changes being undetectable, wouldn't they show up in `terraform plan`?

rfcs/0000-terraform-workflow-improvements.md
```diff
@@ -0,0 +1,856 @@
...
+
+_Modules as immutable artifacts_
+
+Nearly all of CTD's modules are currently stored in the devops repo and called using a relative path to the module's directory. Terraform supports a number of alternate methods for sourcing modules, including:
+
+- Git repositories (including specific refs)
+- Public module registries
+- Private module registries
+
+The main advantage of using a repo or module registry instead of a path is that the module can then be treated as a versioned artifact. If we are able to pin module calls to a specific version, we can make changes to the module without immediately impacting all the contexts in which it's called. This is useful for both infrastructure component modules, which are used across many different applications, and application modules, which are used to build multiple distinct environments. It also reduces the need to rely on feature gating mechanisms such as module variables, which is the typical approach to module changes currently.
```
beatrixada: I agree that this is a potential advantage, but it's also a tradeoff in that it sets up another queue of version upgrades that need to be handled periodically so that configuration doesn't end up relying on a stale version when an urgent fix is made. I think it makes more sense to version the widely-shared standard modules than application modules, but even then it still raises this issue.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: Your assessment is correct that versioning modules will increase the number of pull requests required to apply a change to a running environment. The extra step is necessary in order to ensure adequate change control in production, as described [here](https://github.com/mbta/technology-docs/pull/16/files/57ee6d305eddb5699a5017775357d969199da182#diff-f30d4f210d145fc9215534dd17062f685f7bc5a6f4522b4ca6b3ef7b0e39f9faR160). Outside of production, module versions won't be required, as noted [here](https://github.com/mbta/technology-docs/pull/16/files/57ee6d305eddb5699a5017775357d969199da182#diff-f30d4f210d145fc9215534dd17062f685f7bc5a6f4522b4ca6b3ef7b0e39f9faR444).

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: Changes to child modules will show up in a `terraform plan` in the module where that child module is called, yes. The problem is that, when merging changes to a child module, there's no way to know where it's being called, and thus no way to know whether or not production resources are being impacted by the change.

If we're running `terraform plan` and `terraform apply` in an automated manner (as is proposed elsewhere in this RFC), then we need to be sure that **every** plan and apply contains only the changes we desire. There is no way to guarantee that without versioning child modules in production.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I don't think it's desirable to tightly couple infrastructure changes to application changes in that way. We should discuss this more as a group outside of this RFC, since it seems to indicate a difference in philosophy around how application infrastructure is managed.

beatrixada: I mostly agree, but the way we work now with application configuration specified in Terraform, the coupling is sort of inevitable.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: The existing process requires the Infrastructure team to approve every pull request. Migrating application infrastructure to team-specific _root modules_ (not repos), and allowing team members to approve each others' PRs, will effectively remove the Infrastructure team from the process (with the exception of [changes impacting IAM](https://github.com/mbta/technology-docs/pull/16/files/57ee6d305eddb5699a5017775357d969199da182#diff-f30d4f210d145fc9215534dd17062f685f7bc5a6f4522b4ca6b3ef7b0e39f9faR542-R550)). That should reduce the need to coordinate changes with the Infrastructure team.

Yes, a version bump might constitute an extra PR, but your team will have control over when and how that PR gets approved, merged, and applied. I get your concern that this amounts to extra work, but I feel feel that it's an absolutely necessary layer of protection on production systems, especially if we delegate production apply permissions to a third-party system. That requirement wouldn't change if we moved root modules to application repos.

To your other points:

- Yes, each root module will pave a separate Scalr workspace with [separate permissions](https://github.com/mbta/technology-docs/pull/16/files/57ee6d305eddb5699a5017775357d969199da182#diff-f30d4f210d145fc9215534dd17062f685f7bc5a6f4522b4ca6b3ef7b0e39f9faR530-R540)
- Yes, the Scalr structure and the code review process would be the same regardless of what repo the configuration lives in

I don't mean to imply that the administrative/security implications of moving root modules to `n` repos mean we can't or shouldn't consider it in the future. I'm saying that it's not in the scope of this RFC to think through those implications. And I don't think including it or not has a major impact on the overall workflow being proposed, given the way our applications and infrastructure are currently deployed.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I created a task for the tech leads to discuss alternative application configuration approaches.

rfcs/0000-terraform-workflow-improvements.md
ianwestcott: I think what you're calling out is important knowledge to have when using Terraform, but on a more fundamental level than this RFC aims to cover. I do think that anyone writing Terraform configuration needs to understand the concept of child modules and root modules, and the relationship between the two. Terraform [offers a tutorial](https://developer.hashicorp.com/terraform/tutorials/modules) that covers these concepts.

I added a note about the assumptions this document makes about knowledge of Terraform's concepts: https://github.com/mbta/technology-docs/pull/16/commits/37e3ddb77b797bd0abab685b2f34a0859a08d316

ianwestcott:
I realize now that I failed to note the "motion for final comment period" (FCP) in this thread, instead making the equivalent announcement in Slack on February 10th (and again on the 16th and 21st). This has received a lot of good feedback and generated good discussion, and I'm comfortable moving forward with the recommendations, granting that some things will still need to be worked out along the way.

PR #19, opened 2023-03-21
RFC 19: Event Driven Architecture

paulswartz:
[Rendered version](https://github.com/mbta/technology-docs/blob/ps-eda/rfcs/0019-event-driven-architecture.md)

Comments are open through 2023-03-28.


rfcs/0019-event-driven-architecture.md
```diff
@@ -0,0 +1,183 @@
...
+
+Events which refer to users or operators may contain information about them: names, badge numbers, email addresses, or phone numbers (for example). If the event stream does contain PII, the Kinesis stream MUST be encrypted at rest. Consumers using the stream MUST take care that any PII they receive is properly stored: either ignored, or also encrypted. 
+
+## Idempotent events
+
+Due to the vagaries of the network, events may be sent to the network multiple times. Clients MUST be able to handle receiving an event multiple times. Producers SHOULD use the same  unique identifier (`id` in the CloudEvent) when re-sending an event to make it easier for clients to de-duplicate messages.
+
+## Long-term Storage
+
+As Kinesis only stores events for 24 hours by default, producers SHOULD record events into durable storage as well. [Kinesis Firehose](https://aws.amazon.com/kinesis/data-firehose/) is an AWS service which can perform this role, but is not the only approach.
```
skyqrose: I don't understand this line. Should producers save the events as they're going out, in addition to LAMP saving them as a consumer? Does this just mean that producers should have a real database for their data and shouldn't depend on using Kinesis as de-facto semi-persistent storage? Did you mean that consumers should save the data and not depend on getting past events from Kinesis?

skyqrose (approved):
👍 I like this. It's accurate to what we've been planning to do, and a good summary for people making future streams.

rfcs/0019-event-driven-architecture.md
paulswartz: Great questions! Kinesis, with its limited time horizon, is not persistent storage. Lamp is a destination for events, but not a method of durably storing the events: it could be used for performance-related events. A non-Firehose option would be storing the data in an existing RDS database.

paulswartz:
Marking this as Final Comment Period, with disposition to Accept; comments are still accepted through EOD 2023-04-05.

PR #25, opened 2024-06-05
rfc: Consistent Light Rail Trip Model Across TID Applications

lemald:
[Rendered](https://github.com/mbta/technology-docs/blob/lem-light-rail-trip-model/rfcs/0000-light-rail-trip-model.md)


rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+Of the two models, the proposed end state on light rail most closely resembles the current state on heavy rail. We currently lack the on-vehicle hardware to implement a bus-like solution for light rail, but Glides already exists and provides a means for other operations personnel to provide us with information about which trips are being operated by what vehicles.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+For the purposes of this RFC, we have had to assume that there will still be planned disruptions that have not been modeled in HASTUS in a way that reflects actual operations and trainsheets. This leaves open questions as to whether this will always be the case, and whether or not we will one day be able to assume that the schedule from HASTUS reflects planned operations in all cases.
```
jfabi: Regardless, it'll be the case that one day (actually, [see just two days ago](https://twitter.com/MBTA/status/1797700348276281529)) when an unplanned diversion occurs for which modelled trainsheets won't be available. During this type of situation, operators may be unexpectedly cut off/kept to one half of the line, and operator cycle times may wildly vary compared to schedule.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. Glides may not be aware of some disruptions in its schedule data that are modeled at the `gtfs_creator` level, meaning that we will still need RTR to create ADDED trips for any scheduled trips that it doesn't recognize. With the current logic, RTR will at least attempt to match to trips in the disrupted GTFS. However, given that these trips aren't guaranteed to match actual operational practice, it's unclear how much value there is in matching to them anyway.
+1. It introduces a very slight lag in trip assignments making their way into passenger-facing data when updated due to train movements. RTR will need to receive the updated location information and Glides will need to receive it, and then Glides will in turn need to make the assignment and transit that information to RTR via an event. However, this added delay is likely minimal, especially if we move towards an event-based architecture for sending vehicle position updates from RTR to Glides (out of scope for this RFC but also something I want to consider in the future).
```
jfabi: Flagging for Transit Data that it'll be worth considering which parts of RTR to refactor first (eg, how this piece fits in).

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. This mirrors the situation on heavy rail where the OCS performs realtime trip assignments and RTR uses those as its main source of input for heavy rail trip assignments.
+
+The main downsides of putting this logic in Glides are:
+
+1. As Glides will be more directly responsible for data that flows more directly into passenger-facing systems, it is possible that the Glides team will have to respond to more feedback about issues related to why Green Line trains were assigned to incorrect trips. However, this is likely going to happen anyway with trainsheet-based predictions in RTR given that that already relies on input from Glides.
+1. Glides does not currently have a way of getting location data on yard pull-out / pull-back trains from RTR, so we will need to set up a pipeline for that in order for Glides to implement the full functionality. However, we already need to do this to enable other functionality in Glides, so it is not introducing net new work.
```
jfabi: Interesting...this could come through the existing TMOV pipeline if certain configuration changes are done in coordination with (or by) the OCS team, though it certainly could be done another way (eg, looking at raw GPS records)...

lemald: To be clear, I mean this mainly around mainline moves to / from Inner Belt, which as I understand it RTR currently suppresses from the feed but does receive TMOVs for (for the portion of the trip on the mainline as opposed to in the yard). Tracking trains in yards / off the mainline is another interesting question but I think somewhat orthogonal to this RFC.

rfcs/0000-light-rail-trip-model.md
lemald: Yes, unplanned disruptions are an important consideration that I haven't really addressed explicitly in this RFC. My hope is that edits to trainsheets and things like dropped and added trips would flow through smoothly and RTR would at least be able to use that information effectively. That said, I did have a literal shower thought last night that, while I discuss trip to trip linking for scheduled trips that exist in HASTUS and future TODS, I don't think we really have a way to link added trips aside from via car numbers if we choose to go that route, so that's worth considering.

rfcs/0000-light-rail-trip-model.md
jfabi: Ahh, got it, yes!—hopefully there's an opportunity to work on that in tandem with intaking revenue status from Glides. 🙂

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. Creating clear understanding between teams about ownership of different parts of the data pipeline.
+1. Balancing the needs of internal operations users and public-facing users.
+1. Staying as close as possible to the most basic common denominator in data across TID systems, in this case GTFS trips. This means that improvements benefit as many teams and applications as possible, and we adhere to the principle of "small pieces, loosely joined" by a common standard.
+
+The TID data architecture does not exist in a vacuum. For our purposes in this RFC, the main outside factor that we need to be aware of is the HASTUS system used for scheduling and the Plans & Schedules team that uses it to produce the new light rail schedule every rating. Challenges in this area, in particular around schedules during disruptions, will be factored in.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The below diagram gives an overall view of the proposal. Solid lines represent static data, whereas dotted lines represent realtime data:
```
beatrixada: question: shouldn't there also be a static arrow for Trips from HASTUS to OCS?

lemald: That arrow does exist for heavy rail but I didn't include it here because OCS doesn't handle light rail trainsheets.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+	glides -. Trainsheet Edits .-> rtr
+	ocs -. Vehicle Locations .-> rtr
+	rtr -. Vehicle Locations .-> glides
+	rtr -. Publishes .-> predictions
+	rtr -. Publishes .-> vehicles
+```
+
+## Realtime data: Glides trainsheets and RTR
+
+In the first stage, RTR and Glides will continue to maintain their own separate trip assignment systems, but Glides will begin sending HASTUS-based trip IDs in trainsheet update events. For scheduled trips, the [trip key published in `glides.trips_updated` events](https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.trips_updated.v1#tripkey) will need to be updated in a new version of the schema to include this ID. The `serviceDate` will still need to be present in order to map from HASTUS schedule IDs to GTFS service IDs. The [Glides code that generates the trip key for the event](https://github.com/mbta/glides/blob/fbe9578c8400b955304c74d34e672460a0d17a23/lib/glides_web/channels/trainsheet_channel.ex#L1324) already has access to the `ScheduledTrip` with its trip ID, so once we are importing the trip ID from HASTUS we will simply be able to reference that. This will be a backwards-incompatible change, requiring a new version of the schema.
```
skyqrose: Does adding the hastus id to the [TripKey schema](https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.trips_updated.v1#tripkey) have to be a backwards-incompatible change?

Status Quo:
```json
{
  "serviceDate": "2023-01-20",
  "startLocation": { "gtfsId": "place-mdftf" },
  "endLocation": { "gtfsId": "place-heath" },
  "startTime": "04:47:00",
  "endTime": "05:34:00"
}
```

Currently proposed, backwards-incompatible change:
```json
{
  "serviceDate": "2023-01-20",
  "tripId": "60570849"
}
```

Instead, we could keep the old location/time fields. This would make it a backwards compatible change, and also might be useful for RTR for a fallback if they don't recognize the trip id (because of a disruption).
```json
{
  "serviceDate": "2023-01-20",
  "tripId": "60570849",
  "startLocation": { "gtfsId": "place-mdftf" },
  "endLocation": { "gtfsId": "place-heath" },
  "startTime": "04:47:00",
  "endTime": "05:34:00"
}
```

(Note that the added version of TripKey is unchanged and not mentioned here.)

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Glides schedule import
+
+Glides will import schedule data from either its own HASTUS export or, in the future, TODS. In either case, trip data is loaded into the `scheduled_trips` table. The HASTUS trip ID will be used directly to populate the `trip_id` column. While working directly with HASTUS, some modifications will be necessary to which IDs we actually use. Currently, our export gets a column called `trp_number`, but the actual, database-unique ID is called `trp_int_number`, and the latter is also the value published by `gtfs_creator`. The significant different here is that cars grouped into one logical trip all have the same `trp_number` generally, but as `trp_int_number` is database-unique, that value will differ. We want the benefits of both combined trip information for multiple cars as well as uniqueness and agreement between systems. To achieve this, we will combine the trips for the different constituent cars and determine the ultimate ID in the following manner:
```
skyqrose: An alternative: What if Glides instead kept using its own trip id for its own uses, but saved the hastus trip id in a separate column, and then used that column when publishing?

This would give Glides more resiliency against weird stuff like duplicate hastus ids, and would mean we don't have to mix hastus ids and our own added trip ids in the same column.

I guess this is a little bit like the reason to not use natural keys as a primary key in a DB. Though maybe that analogy implies that we should just skip the Glides-generated trip id and always use either the DB-generated primary key or the natural hastus id.

(This is purely an internal Glides implementation detail, so I guess doesn't really need to be solved in this RFC.)

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+1. Group the trip entries for each schedule ID by the `trp_number` value.
+1. For each resulting group, take the trip record for the lead car and use its `trp_int_number`.
+1. Use information about the trailer trip for data like the run number, but discard its `trp_int_number`.
+
+## Glides trainsheet edit events
+
+The `com.mbta.ctd.glides.trips_updated` event schema will receive a version bump, modifying the `TripKey` datatype. `TripKey` will be the same for both added and scheduled trips, in either case being represented as an object with `"serviceDate"` and `"tripId"` keys, where the `tripId` is sourced from the `trip_id` column of `scheduled_trips` for a scheduled trip, or the auto-generated Glides ID for an added trip.
```
skyqrose: Do consumers need to know the difference between a tripId that comes from HASTUS and one that Glides made up for an added trip? It could matter if they need to know whether they should expect the trip id to be anywhere in GTFS.

If they do need to know the difference, we could add a new `"addedByGlides": bool` (or enum) field in the new TripKey, or we could keep using the old two-case TripKey (which could potentially avoid the backwards-incompatible change).

I guess one downside to keeping the old two-case TripKey is that any other things that reference a trip id (like the new trip_assignment event) need a separate way to say whether it's added or scheduled.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. AVI code (indicating train destination)
+
+While there are not hard requirements around this, Glides should generally try to produce trip assignments for any vehicle tracking in the realtime data that is operating on the mainline (as opposed to in a yard). This is important for RTR to be able to make predictions without doing additional trip assignment logic (any train operating in service should be getting predictions, whether revenue train predictions in the public GTFS-rt feed or non-revenue predictions for in-station announcements about trains that won't stop). Additionally, it has benefits for operations (having a trip in the internal Glides data model provides a location to store any additional metadata, such as notes). For the time being, trip matches performed to provide a placeholder trip for a train not associated with a trip explicitly via the cars entered on a trainsheet will generally be ADDED trips that do not appear on the trainsheet interface. They will, however, generate corresponding `com.mbta.ctd.glides.trips_updated` events creating those added trips.
```
skyqrose: For automatically added trips, we probably can't guarantee that a `trip_updated` event gets received before a `trip_assignment` event, so consumers should be able to handle a trip_assignment that references a never-before-seen trip id anyway. (Though you basically say that in the next paragraph anyway.)

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+Once this feed is implemented, RTR should adopt it as its source of truth for determining trip assignments in the data that it publishes. If RTR receives a `vehicle_trip_assignment` event for a trip ID that it does not recognize (via either the GTFS schedule or via a trip add `trips_updated` event), it should map that to a unique added trip ID of its own and maintain that mapping in the GTFS-rt feed for the duration of the trip (that is, the unrecognized trip ID from Glides should not suddenly become mapped to a different randomly-generated added trip). This is to handle the case where RTR is aware of a different set of trip IDs due to a disruption modeled in `gtfs_creator` but not in HASTUS. See [Rationale and alternatives](#rationale-and-alternatives) for some brief discussion of the tradeoff involved here.
```
skyqrose: Is there a reason not to use the unrecognized Glides-provided id (whether it's added in Glides or came from HASTUS) in GTFS-RT? Why does RTR need to make its own ids and maintain a mapping?

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Glides data requirements
+
+Having Glides handle all trip assignments will require Glides having a direct feed of all vehicle position data. Currently, RTR excludes some known non-revenue yard pull-outs and pull-backs, which makes sense from a passenger information perspective. Internal consumers like Glides will need a more direct feed of the data before any filtering is applied. While the exact mechanism for achieving this is out of scope for this RFC, it is worth noting as a requirement as it relates to other discussions around re-architecting RTR. One possibility that has been discussed is splitting RTR into a data aggregation and state tracking component and a predictions component, set up such that other applications like Glides can also be consumers of the data from the former.
```
skyqrose: Oh, this is a good point that we should highlight: In order for Glides to take over trip matching, we need internal realtime positions.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Glides trainsheet edit events
+
+The `com.mbta.ctd.glides.trips_updated` event schema will receive a version bump, modifying the `TripKey` datatype. `TripKey` will be the same for both added and scheduled trips, in either case being represented as an object with `"serviceDate"` and `"tripId"` keys, where the `tripId` is sourced from the `trip_id` column of `scheduled_trips` for a scheduled trip, or the auto-generated Glides ID for an added trip.
+
+## Glides train - trip assignment events
+
+A new `com.mbta.ctd.glides.vehicle_trip_assignment` event will be established. The exact schema will be defined via the pull request review procedure on the [`mbta/schemas`](https://github.com/mbta/schemas/) repository, but there are some basic criteria that can be established. Events will at least have a `tripId` and `vehicleId` key. The `tripId` will be the scheduled or added trip ID from Glides, and the `vehicleId` will correspond to the ID of the vehicle in the GTFS-rt feed. Each event will represent the assignment of the given `vehicleId` to the given `tripId`. Either a variation of this event or a separate event will exist for unassigning a vehicle from a trip.
```
skyqrose: Should the `trip_assignment` event contain any information besides a tripid<->vehicle mapping? E.g. should it contain the next stop, or the destination, or whehter it's in service? Or should all that come from the existing `trip_update` stream?

It might be nice if users can get some realtime information only from this feed without having to cross reference to another stream. Predictions will definitely require reading the other stream anyway, but maybe other uses wouldn't.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Alternative event schema approaches
+
+The approach outlined to changing the realtime event schemas leans strongly towards maintaining an incremental state transfer approach rather than sending all of the data on a given trip with each update. The rationale for this is that RTR needs to take action to update its own state when it receives information from Glides, and sending a full set of data actually makes it harder for RTR to tell exactly what has changed. It is also conceptually similar to the current OCS message architecture, so maintains a sort of symmetry between different data sources that RTR handles. This should, however, be flagged as an item for discussion in case these assumptions are incorrect.
```
skyqrose: Is this referring to `trip_updated v1`, `trip_updated v2`, or `trip_assignment`?

skyqrose (commented):
To summarize the big steps here:
1. Glides publishes HASTUS trip ids (which match GTFS) in our events stream. RTR probably uses them, but can still do its own assignments.
2. Glides gets internal realtime locations (not discussed much, but a prereq for 3.)
3. The main goal: Glides publishes trip assignments, RTR doesn't need any complex trip matching logic anymore.

rfcs/0000-light-rail-trip-model.md
lemald: @skyqrose and I had discussed this previously and personally I'm open minded to either approach. I think partially this comes down to what Transit Data thinks as the consumer (cc @bklebe since you've already been taking a look).

rfcs/0000-light-rail-trip-model.md
lemald: Well, it's a purely internal implementation detail, but it is also one that I waded into in the RFC, so I think it's fair game. 🙂 I kind of like the idea of having as few different keys as possible to refer to the same thing, which is part of why I proposed simply replacing the old one. Regarding weird stuff like duplicates, keep in mind that we're talking about `trp_int_number` here, which _is_ unique in HASTUS, unlike `trp_number`.

rfcs/0000-light-rail-trip-model.md
lemald: I want to get rid of the two-case TripKey but I agree that an `addedByGlides` boolean is a good idea even if we don't have an immediate need for it.

rfcs/0000-light-rail-trip-model.md
lemald: Yes, that's a good point and one that I should maybe be a bit more explicit about.

rfcs/0000-light-rail-trip-model.md
lemald: My worry is possible confusion if we publish a HASTUS provided ID in the GTFS-rt it could possibly also be the GTFS ID of a trip from a different service (i.e. one that operates when the current disruption isn't happening and instead the base schedule is operating), so it's potentially confusing even if _technically_ it doesn't collide with trip IDs in currently active services.

rfcs/0000-light-rail-trip-model.md
lemald: I think this is another one where my opinion depends a bit on what our consumers think - @bklebe, @nlwstein, and others, any strong opinions?

rfcs/0000-light-rail-trip-model.md
lemald: It's kind of referencing all of them, but that could definitely be made clearer.

lemald: ffc0636

rfcs/0000-light-rail-trip-model.md
lemald: I think this is also a point in favor of doing the staged approach where Glides starts sending HASTUS-based trip IDs before fully taking over trip matching, just because fully taking over trip matching is dependent on other architectural changes outside of the scope of this RFC.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+In the future, Glides will move to importing scheduled trips from the [Transit Operational Data Standard (TODS)](https://ods.calitp.org/) file produced by Transit Data. This will similarly use trip IDs sourced from HASTUS and so the trip ID logic will not have to change, although the `schedule_id` we currently use will likely need to be replaced with a `service_id` to match the GTFS-based TODS standard. This change should be largely orthogonal to other changes proposed in this document, so work on TODS does not need to block adoption of these recommendations, or vice versa.
```
paulswartz: suggestion: since this doesn't block the RFC, I think it would make more sense if it lived at the bottom in the "Future possibilities" section.

rfcs/0000-light-rail-trip-model.md
paulswartz: I would also prefer the backwards-compatible approach that @bklebe suggests.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+Additionally, RTR will need to handle cases where current service as reflected in GTFS static does not match the HASTUS-based trainsheets in Glides, due to planned disruptions modeled by `gtfs_creator`. In these cases, the data will manifest as trainsheet update messages with HASTUS-based trip IDs that do not match up with the trip IDs currently running per the GTFS static schedule. This will continue to be a potential issue, even with trip matching handled in Glides, so it is explored in more depth in the [Reference-level explanation](#reference-level-explanation) section.
```
paulswartz: thought: even in the case where there's a disruption, I believe that the original trip ID is still present as a prefix, so it would still be helpful in matching.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+This intermediate state is proposed to allow for a more incremental implementation that's easier to deliver and provides some value without requiring too much deep refactoring of RTR. It can optionally be skipped if the Glides and Transit Data teams collectively decide that it creates more net work for not enough additional benefit.
+
+## Train - trip assignment feed
```
paulswartz: suggestion: this could also move to the "Future possibilities" section.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+In a future state, Glides will be responsible for light rail trip assignments, and generate a new event feed of train-to-trip assignment events. RTR will use this as its primary source of information 
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## GTFS and TODS schedule data
+
+Light rail schedule data in GTFS and (eventually) TODS files will be produced by `gtfs_creator` based on an export from HASTUS. The HASTUS trips for the lead car will be the basis for GTFS trips, and GTFS trips will take on their IDs. Integrating the schedule for multi-car trips (that is, trips with a trailer) into the TODS data will be defined in a separate discussion about the adoption of TODS. When `gtfs_creator` takes a trip from HASTUS and outputs it with only minor factual corrections, it should retain the original trip ID. However, if a trip is actively modified as part of the `gtfs_creator` disruption modeling process, its ID must be changed so as to not overlap with the HASTUS trip IDs.
```
paulswartz: question: is the `must` in the last sentence a MUST, in that it requires Transit Data to always modify the trip ID if the trip is changing?

rfcs/0000-light-rail-trip-model.md
paulswartz: thought: The staged approach would also allow us to compare the Glides and RTR trip matching, while not making any changes to the public data.

rfcs/0000-light-rail-trip-model.md
mmtchong: In this future, would Glides get schedule data quarterly from TODS and still get personnel data from HASTUS? What if we were able to get HASTUS daily access? Would using the HASTUS-provided trip ID make it easier for us to use HASTUS daily to automatically update operators?

lemald: > In this future, would Glides get schedule data quarterly from TODS and still get personnel data from HASTUS?

Yes, I believe that is correct.

> What if we were able to get HASTUS daily access? Would using the HASTUS-provided trip ID make it easier for us to use HASTUS daily to automatically update operators?

I hadn't actually considered this but it's an interesting angle! Without knowing more about HASTUS daily I'm honestly not sure, though.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Glides schedule import
+
+Glides will import schedule data from either its own HASTUS export or, in the future, TODS. In either case, trip data is loaded into the `scheduled_trips` table. The HASTUS trip ID will be used directly to populate the `trip_id` column. While working directly with HASTUS, some modifications will be necessary to which IDs we actually use. Currently, our export gets a column called `trp_number`, but the actual, database-unique ID is called `trp_int_number`, and the latter is also the value published by `gtfs_creator`. The significant different here is that cars grouped into one logical trip all have the same `trp_number` generally, but as `trp_int_number` is database-unique, that value will differ. We want the benefits of both combined trip information for multiple cars as well as uniqueness and agreement between systems. To achieve this, we will combine the trips for the different constituent cars and determine the ultimate ID in the following manner:
+
+1. Group the trip entries for each schedule ID by the `trp_number` value.
+1. For each resulting group, take the trip record for the lead car and use its `trp_int_number`.
```
paulswartz: question: is this going to generate the right trip ID when the trip reverses, and the trailing car becomes the leading car?

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+The primary drawback of doing this is if we decide that the benefits outlined in the [Motivation](#motivation) section are not worth the effort. There are also potential downsides to having Glides become the source of truth for trip assignment data; some of those are discussed in the [Rationale and alternatives](#rationale-and-alternatives) section below.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Glides versus RTR as the source of truth for assignments
+
+In deciding which system would be the source of truth for assignments between trains and trips, I had to weight the pros and cons of both RTR and Glides taking that logic on, ultimately landing on having it live in Glides. The following factors influenced my decision:
+
+1. Assigning trips to trains is a fundamentally stateful exercise that depends on a trains progress on its previous trip, location, trainsheet data about upcoming trips, and other factors. RTR has some stateful logic, but its state storage and management system is not very advanced. Glides has a relational database that is already updated with information about train movements and trips as they happen.
```
paulswartz: thought: it also means there’s a person in the loop on the Glides side, making decisions about mapping vehicles to trips. that’s true on the bus and (to and extent) on the heavy rail side as well.

rfcs/0000-light-rail-trip-model.md
lemald: That is my understanding as well, though I tend to worry about logic that tries to parse out details from IDs rather than treating them as atomic values.

rfcs/0000-light-rail-trip-model.md
lemald: This is just about the scheduled data for a single trip in one direction, so there are no car numbers to reverse or anything like that, and each trip in each direction is its own freestanding thing in the data (linked by run / block IDs).

rfcs/0000-light-rail-trip-model.md
lemald: It was @skyqrose who proposed the idea, I just tagged @bklebe for a Transit Data perspective. In any event it sounds like there's an emerging preference for the backwards-compatible approach, I'll just wait for someone from Transit Data to ultimately confirm before changing it.

rfcs/0000-light-rail-trip-model.md
skyqrose: As I read it, this isn't a future possibility, it's phase 2 of this proposal.

rfcs/0000-light-rail-trip-model.md
lemald: I tried to make as much of this RFC as possible agnostic as to whether the work happens before or after the move to TODS, since I think that's largely a separate decision that will depend on priorities from the respective product teams. That said, in this instance I do make it sound sort of like a sequencing thing. I've tried to rephrase slightly in a047603, but let me know if you still think that's confusing and I should just move it to future possibilities instead.

rfcs/0000-light-rail-trip-model.md
lemald: That is what I'm proposing, as in my view the transformations applied by `gtfs_creator` in these cases alter the trip so fundamentally that it should no longer be considered the same. If you'd prefer I could try to adopt all-caps MUST / SHOULD / MAY / etc. throughout the document to be a bit clearer about how normative different statements are intended to be.

paulswartz: I think that using the all-caps makes it easier for the teams which need to make changes to understand what's expected of them.

rfcs/0000-light-rail-trip-model.md
paulswartz: Agree that we shouldn't rely on it, but it's better than nothing.

rfcs/0000-light-rail-trip-model.md
paulswartz: That edit looks good to me.

rfcs/0000-light-rail-trip-model.md
lemald: I took a quick stab at making that switch in 4c3e8e7.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. Because RTR has to guess as to the correct trip to assign to a train, trip IDs published in our predictions are often not very sensible, or we give up on guessing a scheduled trip completely and assign a train to an added trip. As a result, passenger-facing systems such as mbta.com or screens that may wish to display both scheduled and predicted departure information can't always be sure if two trips with different IDs in the schedule versus the realtime data are actually "the same." Showing scheduled and realtime data for the same trip side-by-side is also a common pattern in third-party applications, so when a trip fails to match to the schedule the result is a combination of a scheduled trip without realtime tracking and a realtime arrival prediction without a corresponding scheduled time, creating a confusing experience for users.
+1. Analysis of historical data for performance purposes is also complicated by the difficulty in linking a trip from the schedule with a trip from the (historical) realtime data.
```
mmtchong: Analysis is also currently complicated by the difficulty of linking a `tripUpdate` with a trip using data from LAMP. This makes it harder to answer questions like "How many trip have notes, and what types of notes are they"?

rfcs/0000-light-rail-trip-model.md
Whoops: My soft preference is backward compatible as well. While we aren't producing predictions based on glides in production, we are parsing glides messages, so a breaking change means we have to coordinate rollouts.

rfcs/0000-light-rail-trip-model.md
lemald: Good point - I had had LAMP in mind when writing this but didn't explicitly call it out. I added a note in 1dac211.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+## Train - trip assignment feed
+
+In a future state, Glides will be responsible for light rail trip assignments, and generate a new event feed of train-to-trip assignment events. RTR will use this as its primary source of information 
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## GTFS and TODS schedule data
+
+Light rail schedule data in GTFS and (eventually) TODS files will be produced by `gtfs_creator` based on an export from HASTUS. The HASTUS trips for the lead car will be the basis for GTFS trips, and GTFS trips will take on their IDs. Integrating the schedule for multi-car trips (that is, trips with a trailer) into the TODS data will be defined in a separate discussion about the adoption of TODS. When `gtfs_creator` takes a trip from HASTUS and outputs it with only minor factual corrections, it SHOULD retain the original trip ID. However, if a trip is actively modified as part of the `gtfs_creator` disruption modeling process, its ID MUST be changed so as to not overlap with the HASTUS trip IDs.
```
Whoops: Just a commentary on current state of GTFS creator modifying things. We've got a few different relevant cases:

- HASTUS corrections: These are generally minor edits (moving a stop, changing a time) and are handled by patches that don't change the trip_id
- HASTUS modeled diversions: Most diversions anymore we are receiving a new HASTUS export for. These come with a different (HASTUS assigned) service_id and trip_id from what's in the rating. We don't change these trip_ids/service_ids from what HASTUS has provided.
- GTFS creator-modeled diversions: This will update the service_ids and trip_ids in semi-predictable ways. They will never match the original HASTUS trip. That said, these will sometimes update trip_ids for trips which were checked, but which the diversion did not apply to. Example: if a diversion affects only one branch of the RL we will update the trip_ids for both branches.
- Things that are modeled as diversions in GTFS creator even though they are not: These will update the service_id/trip_ids even though the underlying trip is not modified. Example: we will sometimes run a shuttle at the end of a train trip as an ExtendedShuttle. This will modify the train's trip_id/service_id even though the trip isn't actually modified, as an implementation detail. These are usually combined with HASTUS based diversions (truncate the orange line with a new HASTUS export, and add a shuttle at the end).

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Glides train - trip assignment events
+
+A new `com.mbta.ctd.glides.vehicle_trip_assignment` event will be established. The exact schema will be defined via the pull request review procedure on the [`mbta/schemas`](https://github.com/mbta/schemas/) repository, but there are some basic criteria that can be established. Events will at least have a `tripId` and `vehicleId` key. The `tripId` will be the scheduled or added trip ID from Glides, and the `vehicleId` will correspond to the ID of the vehicle in the GTFS-rt feed. Each event will represent the assignment of the given `vehicleId` to the given `tripId`. Either a variation of this event or a separate event will exist for unassigning a vehicle from a trip.
+
+The Glides application will be primarily responsible for establishing the train - trip matching. While exact details of the logic are beyond the scope of this RFC, factors that are likely to be weighed include:
+
+1. Cars entered for trips on trainsheets
```
Whoops: Does Glides have the information needed to do this? Currently RTR does this by (slightly oversimplified)

1. Parsing the OCS provided [route number](https://github.com/mbta/rtr/blob/master/docs/ocs_messages.md?plain=1#L98) to [determine the terminals](https://github.com/mbta/rtr/blob/master/lib/rtr/patterns/pattern.ex#L74)
2. Using the vehicle's current position to determine what track it's on and therefore, what direction it should be moving in.
3. Use 1 and 2 combined with the current time to guess the GTFS trip it's on (or assign an added trip if we can't match)

Does Glides have enough knowledge of the tracks to do 2? Or is the idea that in the future we'll be getting reliable enough information from operators that we don't have to infer, they'll tell us when assigning a train to a new trip?

Also how should reverse trips be handled?

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. AVI code (indicating train destination)
+
+While there are not hard requirements around this, Glides SHOULD generally try to produce trip assignments for any vehicle tracking in the realtime data that is operating on the mainline (as opposed to in a yard). This is important for RTR to be able to make predictions without doing additional trip assignment logic (any train operating in service should be getting predictions, whether revenue train predictions in the public GTFS-rt feed or non-revenue predictions for in-station announcements about trains that won't stop). Additionally, it has benefits for operations (having a trip in the internal Glides data model provides a location to store any additional metadata, such as notes). For the time being, trip matches performed to provide a placeholder trip for a train not associated with a trip explicitly via the cars entered on a trainsheet will generally be ADDED trips that do not appear on the trainsheet interface. They will, however, generate corresponding `com.mbta.ctd.glides.trips_updated` events creating those added trips.
```
Whoops: I think this partially answers the above question: If an operator doesn't assign a train, the trip will be ADDED, but this still requires detecting when a train begins a new trip and assigning the correct information to it when an operator fails to input the information, which I think requires the process described above.

rfcs/0000-light-rail-trip-model.md
Whoops: The enhanced GTFS-RT feed should include non-revenue vehicle movements. The standard feed will remove them.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,211 @@
...
+1. Glides does not currently have a way of getting location data on yard pull-out / pull-back trains from RTR, so we will need to set up a pipeline for that in order for Glides to implement the full functionality. However, we already need to do this to enable other functionality in Glides, so it is not introducing net new work.
+1. Glides will need to generate automatically-created trips for trains in service that aren't matched to any trip on a trainsheet, logic that is currently implemented in RTR. However, this is likely desirable anyway because Glides will need a place to associate things like trip notes on unmatched trains in the future.
+1. Glides may not be aware of some disruptions in its schedule data that are modeled at the `gtfs_creator` level, meaning that we will still need RTR to create ADDED trips for any scheduled trips that it doesn't recognize. With the current logic, RTR will at least attempt to match to trips in the disrupted GTFS. However, given that these trips aren't guaranteed to match actual operational practice, it's unclear how much value there is in matching to them anyway.
```
Whoops: I think I may be missing something here. Glides train - trip assignment events seemed to describe glides producing GTFS trip ids for trips it can match, and added trips ids for those it can't. Under what circumstances is RTR still responsible for generating it's own trip ids?

Whoops:
One other unresolved question that isn't discussed here (and maybe doesn't need to be), but with Glides handling trip matching, RTR will still need to map OCS train movements to Glides trips for the purposes of generating its predictions. I think this will be easier than trip matching, but probably still has some fun edge cases.

rfcs/0000-light-rail-trip-model.md
lemald: Thanks, this is all helpful context.

One thing to keep in mind is that this is all just for light rail, so while it's not a bad idea to keep the Red Line or Orange Line, we're specifically concerned about the Green and Mattapan Lines here, which I mention just in case there is a marked difference in terms of how often Green Line disruptions are modeled in a particular way versus how frequently they're modeled that way on the heavy rail.

Just curious, regarding the last point, do you have a sense of how important an implementation detail that is? As in, is it something that could easily be changed in the future? If not then that's fine, just trying to get a sense of which cases I should plan for existing in the long term.

rfcs/0000-light-rail-trip-model.md
lemald: Sorry, pedantic but important point, but operators do not input any data into Glides; inspectors do.

Regarding 1, this does mean that Glides will need an internal data feed that includes at least the AVI code information, which we don't currently have. I know there has been some background discussion of how to handle this sort of internal information in our feeds but I will add a note explicitly calling that out in the "Glides data requirements" section below (dcebbf1).

I do agree that 2 is the trickiest part here, and the one where RTR arguably has the biggest advantage over Glides currently. I'll have to think the implications of that over a bit more and get back to you.

lemald: > Also how should reverse trips be handled?

This logic is only about determining current trip assignments; I think figuring out what trip a train will be doing next is still a question of predictions, and one where looking at the schedule can come into play (looking at runs or blocks to determine what a train will do next).

I suppose one important case here is where RTR currently generates an ADDED trip, it will also create a reverse trip if the end station of the ADDED trip is among a certain set of terminals, right? That might be a little bit trickier in this approach, but we might also be able to avoid having as many ADDED trips.

rfcs/0000-light-rail-trip-model.md
lemald: I think my above comments reply to this somewhat, and I'll also add that when a trip in Glides is scheduled, the act of the inspector at the departing terminal entering the car numbers also generally enters the car numbers for the corresponding return trip, so in that case Glides will easily be able to detect that the previous trip has completed and the new trip with the same car numbers is now in progress.

rfcs/0000-light-rail-trip-model.md
Whoops: It's definitely not an important detail, and in fact caused us some mild heartburn recently. It's not exactly an easy change, but that has more to do with how gtfs_creator is structured than anything else. I don't see a problem with changing it for this effort.

rfcs/0000-light-rail-trip-model.md
lemald: This is the relevant section from the "Glides train - trip assignment events" section above:

> If RTR receives a vehicle_trip_assignment event for a trip ID that it does not recognize (via either the GTFS schedule or via a trip add trips_updated event), it SHOULD map that to a unique added trip ID of its own and maintain that mapping in the GTFS-rt feed for the duration of the trip (that is, the unrecognized trip ID from Glides should not suddenly become mapped to a different randomly-generated added trip)

So in this case, RTR generates an ID but it should be noted that it's a bit different from the present case, as RTR is not responsible for the full act of assigning a trip, just the ID. The reason I want to avoid simply passing a trip ID from Glides into the public GTFS-rt feed if it doesn't match a trip in the currently active GTFS schedule is that it may be the ID of a trip that is in the GTFS file bot not active on the current service date due to trip ID changes around shuttle activations, and I don't want downstream consumers to potentially be confused by that.

lemald:
> One other unresolved question that isn't discussed here (and maybe doesn't need to be), but with Glides handling trip matching, RTR will still need to map OCS train movements to Glides trips for the purposes of generating its predictions. I think this will be easier than trip matching, but probably still has some fun edge cases.

My idea is that Glides will simply publish the vehicle ID so that RTR can key off of that, _but_ that still has some unfortunate edge cases when we run into train tracking issues that cause one consist to appear as two separate trains and that sort of thing.

rfcs/0000-light-rail-trip-model.md
Whoops: Understanding how dispatching works is a big blind spot for me. Please keep any pedantic or non-pedantic corrections coming.

So, to try to clarify my question here currently, RTR makes the assumption that if a train does the outbound trip, it's coming back, so if a train heads out from Government Center to Boston College, it'll also come back and will generate predictions with either a GTFS trip_id or an ADDED one as appropriate based on the outbound trip.

If Glides is now handling the trip assignments and matching how/when do we get that vehicle X will be doing reverse (GTFS or added) trip Y?

To my understanding, the current state of affairs is that because Government Center is a Glides terminal and Boston College is not, we'll currently only receive Glides updates for trips departing Boston College. Under this new system, will Glides recognize the train has turned around and tell us it's now assigned to Y, or because it's not a Glides terminal is that an RTR concern? If Glides is doing the reverse assignment, how do we handle the case where what Glides eventually assigns it to differs from what we predicted? Vehicle 1234 departs Government Center on trip X (and is assigned it by Glides). RTR predicts it will turn around and serve trip Y when it's done, but when it actually turns around, Glides tells us it's on trip Z.

rfcs/0000-light-rail-trip-model.md
Whoops: > the act of the inspector at the departing terminal entering the car numbers also generally enters the car numbers for the corresponding return trip, so in that case Glides will easily be able to detect that the previous trip has completed and the new trip with the same car numbers is now in progress.

I think this may at least partially answer my question above, although are we assuming we can reliably get inspectors to start entering car numbers? If not we still have the sad path to consider where you have to do this without inspector help

Whoops:
> My idea is that Glides will simply publish the vehicle ID so that RTR can key off of that, _but_ that still has some unfortunate edge cases when we run into train tracking issues that cause one consist to appear as two separate trains and that sort of thing.

Your right, I was overthinking this. I was thinking about if Glides tells us train 1234 will do trip X, then in 30 minutes do trip Y, but that's not what's being proposed here. Glides will tell us 1234 is doing trip X, then later on at some point tell us it's doing trip Y, so RTR only needs to update a simple assignment (Train 1234 is now doing Y) rather than tracking a set of trips 1234 will do and figure out which one specifically it's doing.

rfcs/0000-light-rail-trip-model.md
Whoops: Got it, I think that makes sense. We should still guard against it, but in practice, how likely is this? Picking on Green-B again, it's currently closed past Babcock, so my assumption would be that Glides, working off the rating, would not recognize any of the dispatched trips as GTFS trips (because they all go to Boston College) and would publish all ADDED trips. Am I thinking about this wrong?

rfcs/0000-light-rail-trip-model.md
lemald: I think you're right that [my reply in the other thread](https://github.com/mbta/technology-docs/pull/25#discussion_r1704540308) answers most of this question.

One more pedantic point that Government Center and BC are swapped in your example of which one has a user entering data in Glides and which one gets reverse predictions from RTR, but other than that it's the right basic idea.

I'm thinking at this from a few different angles now and I think it will actually be useful to break this down into a few different cases:

1. A regularly scheduled trip on the Glides trainsheet, with car numbers entered by the inspector and then also transferred over to act as the car numbers for the return trip by Glides. This is the most common and also easiest case.
2. A trip that isn't in the Glides schedule but is explicitly added to the trainsheet by a Glides user. These can be either one way or two way. If it's a two-way trip it should work just like case 1. If it's a one-way added trip we won't automatically get a trip assignment once the train turns around, but maybe that's actually appropriate if the user only indicated that the trip is one-way, and we should instead wait for the user to explicitly add another train for that consist?
3. A trip that never gets added to the Glides trainsheet at all. This is definitely the worst case in a lot of ways. Part of me wants to say that this should be uncommon enough that it might not be worth handling specially. Our team is also going to be doing some work soon on understanding how often there are trains running around that don't correspond to anything on the Glides trainsheets at all, which should give us a sense of how big (or not) of a problem this is, which might change my mind.

rfcs/0000-light-rail-trip-model.md
lemald: Yes, this still relies on inspector input, although in terms of the quality of the data being entered, the problems are more around making sure that the data is entered in advance of the initial departure. In the BC - Government Center example, if a train departs BC and the inspector doesn't enter the car numbers and so on until the train is actually leaving, that's bad for predictions from BC but in terms of figuring out that that train is then going to turn around and make the following trip from Government Center back, that still gives us plenty of lead time.

rfcs/0000-light-rail-trip-model.md
lemald: Okay, updated the document to take the backwards-compatible approach in fab7583.

rfcs/0000-light-rail-trip-model.md
lemald: > I do agree that 2 is the trickiest part here, and the one where RTR arguably has the biggest advantage over Glides currently. I'll have to think the implications of that over a bit more and get back to you.

I did want to quickly circle back to say that I definitely haven't forgotten about this. I did at least make sure to note this as an important consideration in the body of the document (361ad05). I may decide to make other changes but I want to make sure that the concern is at the very least noted.

lemald: Okay, based on a conversation with @skyqrose I have written up some more thoughts on this matter - long story short I think it's okay, but let me know what you think @Whoops. 2fffccb

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+
+The Glides application will be primarily responsible for establishing the train - trip matching. While exact details of the logic are beyond the scope of this RFC, factors that are likely to be weighed include:
+
+1. Cars entered for trips on trainsheets
+1. Additional mid-line edits to trips by Glides users, however those are represented
+1. Train movement (for instance, marking a train as no longer assigned to a trip when it reaches that trip's endpoint, or assigning it to a different trip if appropriate)
+1. Train location (for instance, marking a train as no longer assigned to a trip if its current location is no longer on the path from the start point to the end point, including being headed in the correct direction)
+1. AVI code (indicating train destination)
+
+### Required properties of Glides-produced trip assignments
```
lemald: @Whoops are there any additional things you want us to call out here in terms of what RTR would need / expect in from the trip matching data?

rfcs/0000-light-rail-trip-model.md
paulswartz: probably a question for @Whoops: how is the next stop/direction ID in the GTFS-RT feed calculated now? If it depends on the trip assignment, then I'm not sure how this would work for Glides.

lemald: It's based on pattern, and I addressed how that is determined in the content of the changes I just pushed.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+
+Once this feed is implemented, RTR MUST adopt it as its source of truth for determining trip assignments in the data that it publishes. If RTR receives a `vehicle_trip_assignment` event for a trip ID that it does not recognize (via either the GTFS schedule or via a trip add `trips_updated` event), it SHOULD map that to a unique added trip ID of its own and maintain that mapping in the GTFS-rt feed for the duration of the trip (that is, the unrecognized trip ID from Glides should not suddenly become mapped to a different randomly-generated added trip). This is to handle the case where RTR is aware of a different set of trip IDs due to a disruption modeled in `gtfs_creator` but not in HASTUS. See [Rationale and alternatives](#rationale-and-alternatives) for some brief discussion of the tradeoff involved here.
```
jfabi: Does this also suggest that RTR should send back to Glides a mapping of the new trip IDs so that, for instance, Glides could keep its records straight in the playback map tool and maybe the ladder? Or maybe it's fine if Glides can just align based on the vehicle ID?

rfcs/0000-light-rail-trip-model.md
jfabi: NB for the Transit Data folks that, iff it's decided that there isn't value in trying to match to scheduled disrupted trips in this case (or even if there is!), we should take the opportunity to consider making use of the `DELETED` ScheduleRelationship (https://gtfs.org/realtime/reference/#enum-schedulerelationship_1).

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+
+One important subtlety to note is that we will be relying on next stop ID as an input for Glides trip matching, but in order to determine this in the first place RTR needs to know what pattern a train is on in order to predict what stop it will visit next based on its predicted routing through the system. This raises an interesting question as to how, if at all, RTR should factor in the assigned trip provided by Glides when determining which pattern to assign to a train. If RTR does use the assigned trip for this purpose, there is at least a risk of a circular relationship between the two data points. Whatever the exact relationship ends up being, RTR should probably not take the trip as the only source of input for making this decision, as there are some known cases like short turning trains in the central subway where we know that a train may update its AVI on short notice and, currently at least, the Glides trainsheets may not necessarily be updated to reflect this change in a timely manner. Exactly where to find the balance between these factors is outside of the scope of this RFC, but it should be noted as a risk factor.
```
jfabi: Following up on [a recent discussion](https://github.com/mbta/technology-docs/pull/25/files#r1711532462 ), is this basically saying that RTR would still be expected to make a "preliminary" trip-trip assignment (for the purpose of knowing where it's going/what its next stop will be)? I guess the question of who/what determines a *trip assignment* need not be fully tied to who/what determines a vehicle's *predicted station stops*, as the latter could be influenced not only by trainsheet input but also the AVI codes and GTFS-alerts (and someone/thing needs to decide what how they get prioritized).

jfabi:
> My idea is that Glides will simply publish the vehicle ID so that RTR can key off of that, _but_ that still has some unfortunate edge cases when we run into train tracking issues that cause one consist to appear as two separate trains and that sort of thing.

Unfortunate edge cases that seem to happen on a regular basis, if #light-rail-train-tracking is any indication: When it comes time to consider the matching logic, we should also treat it as an opportunity to review the more common cases of tracking issues from OCS (ie, third car added, separated cars) and see if there are easy (but operationally sensible) ways to address them with the help of Glides inputs.

rfcs/0000-light-rail-trip-model.md
lemald: Good question - I hadn't really considered playback but I'm not sure that I really know enough about our requirements around that to say one way or another. As far as the ladder goes, my assumption is that since Glides would be keeping track of the vehicle ID - trip matching on its own end, it could use that (unless I'm missing some edge case where we would _only_ have RTR-assigned trip ID on some data that we want to associate with the ladder).

rfcs/0000-light-rail-trip-model.md
lemald: Yes, one of my goals is to at least separate these concerns out a bit more explicitly. As I see it, what trips a train / crew will operate next is a bit more of a predictions issue and something we could probably mainly figure out with things like scheduled block / run data (pending validation of course). That said, as we brainstorm and validate approaches to predictions, I think we should think about whether there's an additional data that Glides could provide in order to make that determination easier - so I don't want to just drop the problem entirely on RTR, but I do want to tease it apart as a separate piece of logic.

lemald:
Update: discussed this with Transit Data and @paulswartz yesterday. Long story short, there was buy-in on adopting this RFC. We discussed various aspects of how this relates to terminal predictions and I generally felt like we were on the same page with respect to agreeing that this RFC represents an improvement in the areas where it explicitly seeks to improve while still being cognizant of related risk areas around things we're working on. I've added on additional qualifying remark in 58a3138 based on our conversation, just emphasizing the important of future architectural decisions about the flow of vehicle position data.

I'm therefore going to make a motion for a final comment period, and tentatively target marking the RFC as "accepted" and merging the PR next Tuesday, September 3 if there are no major objections.

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Glides schedule import
+
+Glides will import schedule data from either its own HASTUS export or, in the future, TODS. In either case, trip data is loaded into the `scheduled_trips` table. The HASTUS trip ID will be used directly to populate the `trip_id` column. While working directly with HASTUS, some modifications will be necessary to which IDs we actually use. Currently, our export gets a column called `trp_number`, but the actual, database-unique ID is called `trp_int_number`, and the latter is also the value published by `gtfs_creator`. The significant different here is that cars grouped into one logical trip all have the same `trp_number` generally, but as `trp_int_number` is database-unique, that value will differ. We want the benefits of both combined trip information for multiple cars as well as uniqueness and agreement between systems. To achieve this, we will combine the trips for the different constituent cars and determine the ultimate ID in the following manner:
+
+1. Group the trip entries for each schedule ID by the `trp_number` value.
```
skyqrose: We've noticed issues with using `trp_number` in the past: There was that one time when some trips were added to the rating at the last minute, and they ended up all having the same `trp_number`. I also think that if a two-car pullout trip has different times (because the pilot has to show up a couple minutes earlier than the trailer), they could end up with different `trp_number`s, even when we want to consider them the same trip (I haven't double checked this).

So, I think that `trp_number` may not be reliable for this purpose. Perhaps the RFC should not say _how_ the cars are grouped together into a single trip, and instead leave that up to the implementer, so they could instead base the groups on time+location if it's more reliable.

This comment does not affect 2. and 3., which are only about what to do with the groups, it doesn't matter how they're formed.

rfcs/0000-light-rail-trip-model.md
skyqrose: [Above](https://github.com/mbta/technology-docs/pull/25/files#diff-bc940bb242edf6f22909059b8aae876481e0ae33b958c05cd99e5f375b91f397R108), it says that phase 1, publishing HASTUS trip ids, would be a backwards-compatible change. (This was an edit made as a result of discussion, and wasn't in the first draft.)

Are you still thinking of doing the backwards-incompatible change here, with a new unified TripKey? If so, would that be in phase 1 or phase 2?

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+1. AVI code (indicating train destination)
+
+### Required properties of Glides-produced trip assignments
+
+While there are not hard requirements around this, Glides SHOULD generally try to produce trip assignments for any vehicle tracking in the realtime data that is operating on the mainline (as opposed to in a yard). This is important for RTR to be able to make predictions without doing additional trip assignment logic (any train operating in service should be getting predictions, whether revenue train predictions in the public GTFS-rt feed or non-revenue predictions for in-station announcements about trains that won't stop). Additionally, it has benefits for operations (having a trip in the internal Glides data model provides a location to store any additional metadata, such as notes). For the time being, trip matches performed to provide a placeholder trip for a train not associated with a trip explicitly via the cars entered on a trainsheet will generally be ADDED trips that do not appear on the trainsheet interface. They will, however, generate corresponding `com.mbta.ctd.glides.trips_updated` events creating those added trips.
```
skyqrose: > Glides SHOULD generally try to produce trip assignments for any vehicle tracking in the realtime data

Does this include trains that are turning around? Should we match them to the next trip even before they depart, or is it okay to leave them unmatched at the terminal (like Glides currently does)?

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Required data for Glides to produce trip assignments
+[required-rtr-data]: #required-rtr-data
+
+Having Glides handle all trip assignments will require Glides having a direct feed of all vehicle position data. Currently, RTR excludes some known non-revenue yard pull-outs and pull-backs, which makes sense from a passenger information perspective. Internal consumers like Glides will need a more direct feed of the data before any filtering is applied, in addition to certain non-public data fields like the train's AVI code. While the exact mechanism for achieving this is out of scope for this RFC, it is worth noting as a requirement as it relates to other discussions around re-architecting RTR. One possibility that has been discussed is splitting RTR into a data aggregation and state tracking component and a predictions component, set up such that other applications like Glides can also be consumers of the data from the former.
+
+For determining train location for the purposes of performing trip matching, existing data from RTR in the form of direction ID and next stop ID for trains will suffice for the time being.
```
skyqrose: Didn't we say that AVI would be required, too?

rfcs/0000-light-rail-trip-model.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. Group the trip entries for each schedule ID by the `trp_number` value.
+1. For each resulting group, take the trip record for the lead car and use its `trp_int_number`.
+1. Use information about the trailer trip for data like the run number, but discard its `trp_int_number`.
+
+## Glides trainsheet edit events
+
+The `com.mbta.ctd.glides.trips_updated` event schema will receive a version bump, modifying the `TripKey` datatype. `TripKey` will be the same for both added and scheduled trips, in either case being represented as an object with `"serviceDate"` and `"tripId"` keys, where the `tripId` is sourced from the `trip_id` column of `scheduled_trips` for a scheduled trip, or the auto-generated Glides ID for an added trip.
+
+## Glides train - trip assignment events
```
skyqrose: One point that we've discussed verbally but is unclear as written is that this feed only represents what's happening at the moment, and doesn't represent any planned or predicted assignments in the future.

skyqrose (commented):
I gave this a top-to-bottom reread. Looks good 👍. I left a couple more small comments.

rfcs/0000-light-rail-trip-model.md
lemald: Ah, that was just a case of making the edit in one place but not another. 16b9efd

rfcs/0000-light-rail-trip-model.md
lemald: I seemed to remember something like that happening in the past but then was having a very hard time digging up an actual example when working on this RFC, which makes me wonder if I actually remembered correctly. I also don't want to overcorrect for something that happened once. I made note of it in 88e2688 though.

rfcs/0000-light-rail-trip-model.md
lemald: It is - that's mentioned on line 160.

rfcs/0000-light-rail-trip-model.md
lemald: I would argue that we should assign the vehicle to its next trip in that circumstance. I don't think that the GTFS-rt spec really says much about the semantics here, but I'll point out that with both bus and heavy rail, vehicles laying over at a terminal get assigned to their next trip, if known, rather than sitting around in an unknown state.

PR #30, opened 2026-01-09
Update LRTP_Metrics_Data_Dictionary.md

cleachmbta:
Updates the LRTP Metrics Data Dictionary to include the following:
- Changes to the criteria for false positive departure detection. Instead of choosing which departure is "false" based on time, do so based on the following criteria:
1. If one trip_id contains the string "ADDED-" and the other doesn't, then the "ADDED-" one should be marked as "false."
2. If neither trip_id contains the string “ADDED-” and one trip has predictions and the other doesn't, then the one with no predictions should be marked as "false."
3. If both trips have the same consist, neither trip_id contains the string “ADDED-”, and both have predictions, then fall back to which one was earlier as a tiebreaker.
- Added tables to look 2 predictions forward and 2 predictions back to incorporate into false departure detection logic.
- Count trips with no associated predictions as failed departures incorporated into the calculation of the denominator for the Continuity metric.
- Address false gaps caused by predictions from after the train had already departed.


cho-jos (approved):
overall LGTM. just curious, is there some place where we can see the queries in a more readable format?

data_analytics/LRTP_Metrics_Data_Dictionary.md
```diff
@@ -124,22 +157,28 @@ Prediction Available for Departure? | Identify whether a trip departure has any
...
+Departure Advance Notice >= Min. Advance Notice | Check whether the first prediction generated for per trip departure was generated at least min. advance notice minutes prior to the trip departure time | Boolean | `IF([Advance Notice (minutes) per Departure]>=[Min. Advance Notice (minutes)]) THEN TRUE ELSE FALSE END`
+Departure Advance Notice <= Min. Advance Notice Min Time | Returns the last time that a prediction was generated per trip if it is generated less than of equal to the min. advance notice | Date & Time | `{FIXED [VehiclePositions Unique Daily Trip Identifier]: MIN(IF [Advance Notice (minutes)]<=[Min. Advance Notice (minutes)] THEN [Prediction Generated Time] END)}`
```
djpowers:
```suggestion
Departure Advance Notice <= Min. Advance Notice Min Time | Returns the last time that a prediction was generated per trip if it is generated less than or equal to the min. advance notice | Date & Time | `{FIXED [VehiclePositions Unique Daily Trip Identifier]: MIN(IF [Advance Notice (minutes)]<=[Min. Advance Notice (minutes)] THEN [Prediction Generated Time] END)}`
```
Looking at the query, possible typo here?

data_analytics/LRTP_Metrics_Data_Dictionary.md
```diff
@@ -124,22 +157,28 @@ Prediction Available for Departure? | Identify whether a trip departure has any
...
+Departure Advance Notice > Min. Advance Notice Max Time | Returns the first time that a prediction was generated per trip if it is generated less than of equal to the min. advance notice | Date & Time | `{FIXED [VehiclePositions Unique Daily Trip Identifier]:  MAX(IF [Advance Notice (minutes)]>[Min. Advance Notice (minutes)] THEN [Prediction Generated Time] END)}`
+Continuous Coverage Between Advance Notice > Min. Advance Notice and Advance Notice <= Min. Advance Notice Predictions | Checks whether there is a gap between the first time the trip departure generates predictions and last time that a prediction was generated per trip  if it is generated less than of equal to the min. advance notice | Boolean | `IF ((DATEDIFF('second',[Departure Advance Notice > Min. Advance Notice Max Time],[Departure Advance Notice <= Min. Advance Notice Min Time]))>20) THEN FALSE ELSE TRUE END`
```
djpowers:
```suggestion
Continuous Coverage Between Advance Notice > Min. Advance Notice and Advance Notice <= Min. Advance Notice Predictions | Checks whether there is a gap between the first time the trip departure generates predictions and last time that a prediction was generated per trip if it is generated less than or equal to the min. advance notice | Boolean | `IF ((DATEDIFF('second',[Departure Advance Notice > Min. Advance Notice Max Time],[Departure Advance Notice <= Min. Advance Notice Min Time]))>20) THEN FALSE ELSE TRUE END`
```

data_analytics/LRTP_Metrics_Data_Dictionary.md
```diff
@@ -124,22 +157,28 @@ Prediction Available for Departure? | Identify whether a trip departure has any
...
+Departure Advance Notice >= Min. Advance Notice | Check whether the first prediction generated for per trip departure was generated at least min. advance notice minutes prior to the trip departure time | Boolean | `IF([Advance Notice (minutes) per Departure]>=[Min. Advance Notice (minutes)]) THEN TRUE ELSE FALSE END`
+Departure Advance Notice <= Min. Advance Notice Min Time | Returns the last time that a prediction was generated per trip if it is generated less than of equal to the min. advance notice | Date & Time | `{FIXED [VehiclePositions Unique Daily Trip Identifier]: MIN(IF [Advance Notice (minutes)]<=[Min. Advance Notice (minutes)] THEN [Prediction Generated Time] END)}`
+Departure Advance Notice > Min. Advance Notice Max Time | Returns the first time that a prediction was generated per trip if it is generated less than of equal to the min. advance notice | Date & Time | `{FIXED [VehiclePositions Unique Daily Trip Identifier]:  MAX(IF [Advance Notice (minutes)]>[Min. Advance Notice (minutes)] THEN [Prediction Generated Time] END)}`
```
djpowers:
```suggestion
Departure Advance Notice > Min. Advance Notice Max Time | Returns the first time that a prediction was generated per trip if it is generated less than the min. advance notice | Date & Time | `{FIXED [VehiclePositions Unique Daily Trip Identifier]:  MAX(IF [Advance Notice (minutes)]>[Min. Advance Notice (minutes)] THEN [Prediction Generated Time] END)}`
```
Based on the name and query, looks like this one does not check equality?

djpowers (changes requested):
Looks good! Just highlighted a few potential fixes on some of the query descriptions.

